PR #3, opened 2019-09-10
Run CI tests for old Zivid versions

apartridge:
Update current Zivid version to 1.5.0, and run build_and_test for old
versions 1.3.0 and 1.4.0 for both melodic and kinetic.

Use azure-pipelines templates to reduce code duplication and make it
easier to specify what versions should be tested against what ROS
distros and compilers.


apartridge:
This has passed the checks, see Check tabs, but for some reason is still showing "Some checks haven’t completed yet"  in the Conversation view. Will try to re-start tests.

apartridge:
For some reason the github PR is expecting 5 build results from outdated targets, these targets are no longer being built after this PR. So I cannot merge. :(

PR #9, opened 2019-12-05
Add support for 2d api

apartridge:
Recommend looking at each commit in isolation. While the PR is pretty big, only the last 2 commits contain big changes.


azure-pipelines.yml
```diff
@@ -6,7 +6,7 @@ schedules:
     - master
   always: true
 variables:
-  newest_zivid_version: '1.5.0+63f281e2-26'
```
apartridge: test comment

zivid_camera/CMakeLists.txt
```diff
@@ -30,21 +30,9 @@ find_package(catkin REQUIRED COMPONENTS
...
-
-if((NOT Zivid_VERSION_MAJOR EQUAL SUPPORTED_ZIVID_MAJOR_VERSION) OR
-   (Zivid_VERSION VERSION_LESS MIN_SUPPORTED_ZIVID_VERSION))
-  MATH(EXPR SUPPORTED_MAJOR_VERSION_PLUS_ONE "${SUPPORTED_ZIVID_MAJOR_VERSION}+1")
-  message(FATAL_ERROR "Zivid version ${Zivid_VERSION} is not supported by this ROS driver! \
-    Zivid ${MIN_SUPPORTED_ZIVID_VERSION} or higher, and less than \
-    ${SUPPORTED_MAJOR_VERSION_PLUS_ONE}.0.0, is required. Hint: Download/install the \
-    newest versions of the Zivid ROS driver and the Zivid SDK.")
-endif()
-
```
eskaur: I see you removed a lot of the error feedback. What kind of error message am I supposed to get now if I try to build with SDK 1.5 installed?
I get the rather cryptic:
`./zivid_camera_settings_generator: symbol lookup error: ./zivid_camera_settings_generator: undefined symbol: _ZN5Zivid10Settings2DD1Ev`
which is difficult for the customer if they don't know that Settings2D was released after 1.5.

eskaur: Never mind, I may not have had a completely clean build. I tried from scratch and got this nice one:

```
  Could not find a configuration file for package "Zivid" that is compatible
  with requested version "1.6.0".

  The following configuration files were considered but not accepted:

    /usr/lib/cmake/Zivid/ZividConfig.cmake, version: 1.5.0

```

README.md
```diff
@@ -259,7 +259,7 @@ rosrun rqt_reconfigure rqt_reconfigure
 
 The capture settings available in the `zivid_camera` node matches the settings in the Zivid API.
 To become more familiar with the available settings and what they do, see the API reference for the
-[Settings](http://www.zivid.com/hubfs/softwarefiles/releases/1.2.1+6ca8e224-12/doc/cpp/classZivid_1_1Settings.html)
+[Settings](http://www.zivid.com/hubfs/softwarefiles/releases/1.7.0+a115eaa4-4/doc/cpp/classZivid_1_1Settings.html)
```
eskaur: Do we really not have some URL that always points to the docs for the latest version? This could get annoying when we release so often.

zivid_samples/CMakeLists.txt
```diff
@@ -14,35 +14,55 @@ catkin_package(
...
+
+function(register_python_sample)
+  cmake_parse_arguments(
+    ARG
+    ""
+    "SRC"
+    ""
+    ${ARGN}
+    )
+  install(PROGRAMS
```
runenordmo: I have also seen `catkin_install_python(PROGRAMS scripts/<myScript> DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION})` being used. Are you aware if there is any difference?

README.md
apartridge: We do not have this. I think for now we just have to remember to update these links when we release, especially if we have done changes to Settings. Ideally these links should not go to the API reference but some knowledge base article that describes all the settings, fex https://zivid.atlassian.net/wiki/spaces/ZividKB/pages/99713044/Zivid+One+Camera+Settings, but we don't have good permalinks into the knowledgebase.

zivid_samples/launch/sample.launch
```diff
@@ -1,6 +1,8 @@
 <launch>
+    <arg name="type"/>
+    <arg name="name" default=""/>
```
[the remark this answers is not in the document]
apartridge: When empty string is used, it will end up using a default node name which is specified in the sample itself, like 
`rospy.init_node("sample_capture_2d_py", anonymous=True)`

zivid_camera/src/zivid_camera.cpp
```diff
@@ -58,6 +58,36 @@ std::string toString(zivid_camera::CameraStatus camera_status)
...
+  const auto max_config = getMaxConfigFromZividSettings<ConfigType>(defaultSettings);
+  server->dr_server.setConfigMax(max_config);
+
+  const auto default_config = getDefaultConfigFromZividSettings<ConfigType>(defaultSettings);
+  server->dr_server.setConfigDefault(default_config);
+  server->dr_server.updateConfig(default_config);
+
+  // Setup the callback. This will invoke the callback, ensuring that the locally cached
+  // copy of config is updated.
+  auto cb = [& s = *server](const ConfigType& config, uint32_t /*level*/) {
```
runenordmo: Nice way of not having to duplicate setting up these dynamic reconfigure server callbacks 👍

zivid_samples/CMakeLists.txt
apartridge: Based on /opt/ros/melodic/share/catkin/cmake/catkin_install_python.cmake this seems to do some additional stuff like adding a shebang line which we already have, some stuff for installing it on Python etc. I suggest to leave this as is for now, until we get complaints.

zivid_camera/src/zivid_camera.cpp
```diff
@@ -381,19 +428,14 @@ bool ZividCamera::isConnectedServiceHandler(IsConnected::Request&, IsConnected::
 void ZividCamera::publishFrame(Zivid::Frame&& frame)
 {
   const bool publish_points = points_publisher_.getNumSubscribers() > 0 || use_latched_publisher_for_points_;
```
runenordmo: Suggest extracting to a function `shouldPublishPoints()` for consistency with `shouldPublishColorImg()`

zivid_camera/src/zivid_camera.cpp
```diff
@@ -381,19 +428,14 @@ bool ZividCamera::isConnectedServiceHandler(IsConnected::Request&, IsConnected::
...
-  const bool publish_color_img =
-      color_image_publisher_.getNumSubscribers() > 0 || use_latched_publisher_for_color_image_;
+  const bool publish_color_img = shouldPublishColorImg();
   const bool publish_depth_img =
```
runenordmo: Suggest extracting to a function `shouldPublishDepthImage()` for consistency with `shouldPublishColorImg()`

zivid_camera/src/zivid_camera.cpp
```diff
@@ -420,11 +462,25 @@ void ZividCamera::publishFrame(Zivid::Frame&& frame)
...
 
+bool ZividCamera::shouldPublishColorImg() const
+{
+  return color_image_publisher_.getNumSubscribers() > 0 || use_latched_publisher_for_color_image_;
+}
+
+std_msgs::Header ZividCamera::makeHeader()
+{
+  std_msgs::Header header;
+  header.seq = header_seq_++;
```
runenordmo: I think the sequence number has been deprecated for quite a while, but I its sensible to set it like this anyway.

zivid_camera/src/zivid_camera.cpp
```diff
@@ -475,11 +531,23 @@ sensor_msgs::ImageConstPtr ZividCamera::makeColorImage(const std_msgs::Header& h
   return msg;
 }
 
+sensor_msgs::ImageConstPtr ZividCamera::makeColorImage(const std_msgs::Header& header,
+                                                       const Zivid::Image<Zivid::RGBA8>& image)
+{
+  auto msg = boost::make_shared<sensor_msgs::Image>();
+  fillCommonMsgFields(*msg, header, image.width(), image.height());
+  msg->encoding = sensor_msgs::image_encodings::RGBA8;
+  msg->step = static_cast<uint32_t>(4 * image.width());
```
runenordmo: Suggestion: give a name to the number `4` here. I guess it is the number of bytes corresponding to `R`, `G`, `B` and `A`, but would be nice to have it explicit.

zivid_camera/test/test_zivid_camera.cpp
```diff
@@ -335,6 +349,84 @@ TEST_F(ZividNodeTest, testDynamicReconfigureNodesAreAvailable)
...
+  ASSERT_TRUE(ros::service::call(capture_2d_service_name, capture));
+  sleepAndSpin(ros::Duration(1));
+  assert_num_topics_received(1);
+
+  auto verifyImageAndCameraInfo = [this](const auto& img, const auto& info) {
+    assertCameraInfoForFileCamera(info);
+
+    ASSERT_EQ(img.width, 1920U);
+    ASSERT_EQ(img.height, 1200U);
+    ASSERT_EQ(img.step, 4U * 1920U);
```
runenordmo: Suggestion: give a name to the number 4 here.

zivid_camera/srv/Capture2D.srv
```diff
@@ -0,0 +1 @@
+---
```
runenordmo: Just curious: Why are we not using `std_srvs::Empty` here? Is it because we want the service to be named as what it does? Or is it in case the service will be expanded in the future?

I think it is nice that it is named as what it does, but I guess also there is a trade-off where we want to use as many standard ROS message/service types as possible.

zivid_samples/scripts/sample_capture_2d.py
```diff
@@ -0,0 +1,50 @@
...
+        frame0_config_client = dynamic_reconfigure.client.Client(
+            "/zivid_camera/capture_2d/frame_0"
+        )
+        frame0_config = {
+            "enabled": True,
+            "iris": 27,
+            "exposure_time": 10000,
+            "gain": 1.0,
+            "brightness": 1.0,
+        }
```
eskaur: This is pretty under-exposed for most scenes. May I suggest opening the iris to 35?
(if you agree, remember to also do it with the c++ sample)

zivid_camera/src/zivid_camera.cpp
eskaur: I have already started testing all the samples, so I can deal with that.

zivid_samples/scripts/sample_capture_2d.py
apartridge: Yes can bump it to 35.

zivid_camera/src/zivid_camera.cpp
apartridge: can do

zivid_camera/test/test_zivid_camera.cpp
apartridge: can do

README.md
```diff
@@ -286,84 +296,109 @@ The available capture settings are organized into a hierarchy of configuration n
...
 ## Sample .launch files
 
 ### zivid_camera_with_settings.launch
 
 [zivid_camera_with_settings.launch](./zivid_samples/launch/zivid_camera_with_settings.launch) starts
-the `zivid_camera` node with capture settings set to 3-frame HDR. Modify the settings in this .launch
+the `zivid_camera` node with 3D capture settings set to 3-frame HDR. Modify the settings in this .launch
 file to match your scene.
 
 ```bash
```
eskaur: This is outside the scope of this PR, but: How does one then actually visualize this 3-frame HDR?

I don't know how to combine it with the visualization-instructions below.

zivid_camera/src/zivid_camera.cpp
eskaur: All samples and tests run fine here.

README.md
apartridge: Should work to 
roslaunch zivid_samples zivid_camera_with_settings.launch
ROS_NAMESPACE=zivid_camera roslaunch zivid_camera visualize.launch
then trigger capture
rosservice call /zivid_camera/capture

zivid_camera/srv/Capture2D.srv
apartridge: It is mainly for future proofing. For example, it could be that we wanted to add a return value which can be used to identify which point cloud belongs to which capture.

runenordmo: Yes, I think that is a good reason (in addition to naming the intent of the message). 

The service return value could in theory also be the color image itself. Don’t think we should change that of course, since that is not how we do it for the 3D capture.

apartridge:
Addressed PR comments

zivid_camera/src/zivid_camera.cpp
apartridge: done

apartridge:
Squashed together fixup commits now

runenordmo:
Looks good to me 👍

apartridge:
Any further comments @eskaur?

PR #12, opened 2019-12-10
Add support for capture assistant

apartridge:


eskaur:
For the purposes of giving a good first-impression, I think the first thing the new user tries (under "Getting started") should be a sample that runs the Capture Assistant once (with a relatively small budget) and then loops that capture over and over. 
Agree?

I see that `sample_capture_assistant` almost satisfies this, if it kept capturing beyond just once.

README.md
```diff
@@ -393,6 +448,26 @@ rosrun zivid_samples sample_capture_2d_cpp
 rosrun zivid_samples sample_capture_2d.py
 ```
 
+### Sample Capture Assistant
```
eskaur: This sample doesn't continue capturing in a loop like `sample_capture`. Why not?

README.md
```diff
@@ -208,52 +209,102 @@ ROS_NAMESPACE=zivid_camera rosrun zivid_camera zivid_camera_node _frame_id:=zivi
 
 ## Services
 
-`capture` ([zivid_camera/Capture](./zivid_camera/srv/Capture.srv))
-> Invoke this service to trigger a 3D capture. See section [Configuration](#configuration) for how to
-> configure the 3D capture settings. The resulting point cloud is published on topic `points`, color image
-> is published on topic `color/image_color`, and depth image is published on topic `depth/image_raw`.
-> Camera calibration is published on topics `color/camera_info` and `depth/camera_info`.
+### capture_assistant/suggest_settings
```
eskaur: How exactly would I launch this service from the command line? It's not obvious to me how to specify the two arguments.

Should we add an example-line here to demonstrate?

apartridge: We could add examples of how to launch from command line. Command line is more of a troubleshooting/debugging tool. I dont think it should be added here as it is noise. Maybe under FAQ. 

You can try rosservice call /zivid_camera/capture_assistant/suggest_settings <tab><tab>

apartridge:
> For the purposes of giving a good first-impression, I think the first thing the new user tries (under "Getting started") should be a sample that runs the Capture Assistant once (with a relatively small budget) and then loops that capture over and over.
> Agree?
> 
> I see that `sample_capture_assistant` almost satisfies this, if it kept capturing beyond just once. However, we may want a separate sample if we want quicker captures (smaller budget).

Yes, probably a good idea. As discussed IRL I am not sure if we should loop. Maybe just perform one capture and show it.

README.md
apartridge: As discussed IRL, the reason the other ones are in a loop is to make it easy to change settings and visualize the output. That is not so relevant for this sample. Also, having it in a loop IMO will make it more difficult to understand how the CA works with the initial captures and then the actual capture, fex. observing how long the two parts take.

eskaur: You make good points. I agree now. :)

eskaur:
> > For the purposes of giving a good first-impression, I think the first thing the new user tries (under "Getting started") should be a sample that runs the Capture Assistant once (with a relatively small budget) and then loops that capture over and over.
> > Agree?
> > I see that `sample_capture_assistant` almost satisfies this, if it kept capturing beyond just once. However, we may want a separate sample if we want quicker captures (smaller budget).
> 
> Yes, probably a good idea. As discussed IRL I am not sure if we should loop. Maybe just perform one capture and show it.

I agree. Let the first sample run the capture assistant (1.2s, none-mode), capture once, and show in visualizer.

zivid_camera/src/zivid_camera.cpp
```diff
@@ -379,6 +366,72 @@ bool ZividCamera::capture2DServiceHandler(Capture::Request&, Capture::Response&)
...
+  Zivid::CaptureAssistant::SuggestSettingsParameters suggest_settings_parameters(max_capture_time,
+                                                                                 ambient_light_frequency);
+  ROS_INFO_STREAM("Getting suggested settings using parameters: " << suggest_settings_parameters);
+  const auto suggested_settings{ Zivid::CaptureAssistant::suggestSettings(camera_, suggest_settings_parameters) };
+
+  if (suggested_settings.empty())
+  {
+    throw std::runtime_error("The suggestSettings function returned 0 settings!");
+  }
+  if (suggested_settings.size() > capture_frame_config_dr_servers_.size())
```
runenordmo: It seems a strange to me that we can get an exception here for what might be a quite normal use case (f.ex. the user setting the max number of frames to 3, and the Capture Assistant then suggesting to use 4 frames). 

I guess it will improve with the possibility to specify a maximum number of frames also for the Capture Assistant.

In the mean time, why are we not using the `n=capture_frame_config_dr_servers_.size()) ` frames with highest priority from the Capture Assistant?

zivid_camera/src/settings_generator.cpp
```diff
@@ -305,41 +308,53 @@ class GetConfigMinMaxDefFromZividSettingsGenerator
     }
   }
 
-  std::string typeUcFirst()
+  std::string initializeConfigFunction() const
   {
     switch (type_)
     {
+      case Type::CurrentValue:
+        return "__getDefault__";
```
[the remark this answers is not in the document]
apartridge: This is not our method. This is the name of a auto generated method (which makes some C++ types based on the .cfg files generated in settings_generator.cpp)

zivid_camera/src/zivid_camera.cpp
apartridge: I mention in the readme that the users should not lower the max number of frames parameter from 10, which is the default. And in SDK 1.8 the API should enforce that it does not return more than 10 frames. I have not yet seen more than 9 frames even with max budget. But still the API should not return more than 10 in any situation. Maybe we need to reconsider this in the future if we have a valid use case for more than 10 frames, for example by using a lot of averaging.

I don't think we should silently ignore it here and just take the N highest priority ones. One problem is that I believe the settings from the API are currently not ordered by priority.

zivid_camera/src/zivid_camera.cpp
```diff
@@ -621,4 +644,42 @@ sensor_msgs::CameraInfoConstPtr ZividCamera::makeCameraInfo(const std_msgs::Head
   return msg;
 }
 
+template <typename ConfigType>
+template <typename ZividSettings>
+ZividCamera::ConfigDRServer<ConfigType>::ConfigDRServer(const std::string& name, ros::NodeHandle& nh,
+                                                        const ZividSettings& defaultSettings)
+  : name_(name), dr_server_(dr_server_mutex_, ros::NodeHandle(nh, name_)), config_(ConfigType::__getDefault__())
```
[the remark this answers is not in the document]
apartridge: Same answer as above

zivid_camera/src/zivid_camera.cpp
```diff
@@ -379,6 +366,72 @@ bool ZividCamera::capture2DServiceHandler(Capture::Request&, Capture::Response&)
...
+    if (capture_frame_config_dr_servers_[i]->config().enabled)
+    {
+      ROS_INFO_STREAM("Frame config " << i << " was enabled, so disabling it");
+      auto config = capture_frame_config_dr_servers_[i]->config();
+      config.enabled = false;
+      capture_frame_config_dr_servers_[i]->setConfig(config);
+    }
+  }
+
+  return true;
```
runenordmo: There is no case where this can become false? 
Is there some scenario that it could make sense for the service call to return false, or do you think all these cases should rather be covered by throwing exceptions (as above)?

zivid_camera/srv/CaptureAssistantSuggestSettings.srv
```diff
@@ -0,0 +1,7 @@
+uint8 AMBIENT_LIGHT_FREQUENCY_NONE=0
+uint8 AMBIENT_LIGHT_FREQUENCY_50HZ=1
+uint8 AMBIENT_LIGHT_FREQUENCY_60HZ=2
+
+duration max_capture_time
+uint8 ambient_light_frequency
+---
```
runenordmo: For me it would make sense that the user could get the suggested settings from the Capture Assistant in the service response, and then he could provide the settings to to the dynamic reconfigure server (or to `capture(camera, settings)`).

But I guess the decision to do it like this is based on that we are already using the dynamic reconfigure server for the settings? So that if the user wants to check the output from the Capture Assistant, he will query the dynamic reconfigure server and alternatively disable some of the frames suggested by the Capture Assistant there, if that is desireable?

zivid_camera/src/zivid_camera.cpp
```diff
@@ -379,6 +366,72 @@ bool ZividCamera::capture2DServiceHandler(Capture::Request&, Capture::Response&)
...
+  }
+  if (suggested_settings.size() > capture_frame_config_dr_servers_.size())
+  {
+    throw std::runtime_error("The number of suggested settings (" + std::to_string(suggested_settings.size()) + ") " +
+                             "is larger than the number of dynamic_reconfigure capture/frame_<n> servers (" +
+                             std::to_string(capture_frame_config_dr_servers_.size()) + "). Increase launch parameter " +
+                             "num_capture_frames. See README.md for more information.");
+  }
+
+  ROS_INFO_STREAM("CaptureAssistant::suggestSettings returned " << suggested_settings.size() << " settings");
```
runenordmo: Suggestion: Move this print to the sample, showing how to check what settings were suggested by the Capture Assistant

zivid_samples/src/sample_capture_assistant.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+  CHECK(ros::service::waitForService(ca_suggest_settings_service_name, default_wait_duration));
+
+  ros::AsyncSpinner spinner(1);
+  spinner.start();
+
+  auto points_sub = n.subscribe("/zivid_camera/points", 1, on_points);
+  auto image_color_sub = n.subscribe("/zivid_camera/color/image_color", 1, on_image_color);
+
+  capture_assistant_suggest_settings();
```
runenordmo: Suggestion: Show how to check what settings were suggested by the Capture Assistant, maybe by printing the number of frames suggested.

Similarly for the Python sample.

zivid_camera/src/settings_generator.cpp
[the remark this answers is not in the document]
runenordmo: Ah, I see!

zivid_camera/test/test_zivid_camera.cpp
```diff
@@ -430,6 +447,194 @@ TEST_F(ZividNodeTest, test2DSettingsDynamicReconfigureNodesAreAvailable)
...
+    }
+  }
+}
+
+// TODO(10): Enable this test when minimum Zivid SDK is bumped to 1.8
+TEST_F(ZividCATest, DISABLED_test_going_from_several_frames_to_1_frame)
+{
+  using Request = zivid_camera::CaptureAssistantSuggestSettings::Request;
+  performSuggestSettingsAndCompareWithCppAPI(ros::Duration{ 10.0 }, Request::AMBIENT_LIGHT_FREQUENCY_NONE);
+  ASSERT_GT(numEnabled3DFrames(), 1U);
```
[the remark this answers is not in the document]
apartridge: Yes this testing is based on that the output from the CA running on MiscObjects.zdf is deterministic.

zivid_camera/src/zivid_camera.cpp
apartridge: I think the service should either throw or return true. I don't see a good reason for it to return false.

zivid_camera/src/zivid_camera.cpp
runenordmo: I agree we should not silently ignore it, we could maybe print with `ROS_WARN(...)`. However, I would be suprised as a user if I could not use the Capture Assistant if I set the max number of frames in ROS to 5 and the Capture Assistant suggested 6 frames.

Yes, if the settings from the Capture Assistant are not ordered by priority, we cannot do this.
However, we could set the number of frames / dynamic reconfigure servers in ROS fixed to be 10? Or is there a disadvantage to this?

zivid_camera/srv/CaptureAssistantSuggestSettings.srv
apartridge: We chose to use dynamic_reconfigure for the settings tree. If we do as you propose, then we need a whole system for auto generating the messages to represent the settings tree, which has its own system compared to dynamic_reconfigure. We still need dynamic_reconfigure for having a way to visualize the settings in a GUI, and it is more common to use dynamic_reconfigure for this. I don't see that this gives much value, I would prefer to have one system for simplicity.

zivid_samples/src/sample_capture_assistant.cpp
apartridge: This gets pretty complicated. Would need to generate 10 dynamic_reconfigure clients, then fetch those settings and print them. The sample.launch script already launch rqt_reconfigure, where the settings are visible. The other sample shows how to make a dynamic_reconfigure client, if you want to do manual settings, but it can also be used to read out the settings. I don't think this is worth doing in this sample.

apartridge:
Updated PR to use sample_capture_assistant in Getting started, and made the CA sample the first sample under "Samples".

zivid_camera/src/zivid_camera.cpp
apartridge: It is documented in the README when changing that parameter that it should not be lower than 10 if using CA.

> We do not recommend lowering this setting, especially if you are using the
> [capture_assistant/suggest_settings](#capture_assistantsuggest_settings) service.

Therefore I think it should be exception. Otherwise someone may hit this by accident, for instance if they just copied some .launch file from someplace which happened to have a lower num_capture_frames, and then they get surprised when they get bad data.

The SDK should be fixed to ensure that they are returning settings in prioritized order. @eskaur  do we have an issue on this?

We could maybe have hard coded it to 10, however it could be that for some use case like inspection customer would want to do more than 10 frames, for example doing 20 frames with averaging is a valid use case. So I made it a launch parameter. 

It would maybe be better if this number itself was a dynamic reconfigure setting. So we at runtime can add or remove these nodes as needed. It would probably make client code a bit more tricky. They would need to handle the service possibly going down and coming up again. Not sure how well that works.

kontramind: We have just made PR where max frame limit is hardcoded to 10.
It was our earlier understanding that this is a sufficient number.

kontramind: ATM the returned order is dictated by the exposure value - small exposure values targeting highlights are in the top.

zivid_camera/srv/CaptureAssistantSuggestSettings.srv
kontramind: @runenordmo As discussed IRL, frame importance is different for different applications/users. Some may prefer speed over coverage, and others may want the opposite.

apartridge:
Added commit _test_zivid_camera.cpp: Increase timeout waiting for zivid node to be ready_ which hopefully makes the CI more stable.

zivid_samples/src/sample_capture_assistant.cpp
runenordmo: Yes, you make a good point, also think it is not worth doing in this sample 👍 
Moreover, if someone wanted to go around the dynamic reconfigure option, they would use the API instead. This is also visible how to do it in `zivid_camera.cpp`.

zivid_camera/src/zivid_camera.cpp
runenordmo: We should agree on the maximum frame limit @apartridge @kontramind

apartridge: There is agreement that the SDK will be changed to return max 10 frames. This will work with ROS with default launch parameters, and it is documented in the README that lowering the maximum is not recommended using CA. Maybe the README can be made even clearer in this regard.

zivid_camera/src/zivid_camera.cpp
```diff
@@ -323,27 +326,11 @@ bool ZividCamera::captureServiceHandler(Capture::Request&, Capture::Response&)
...
-      ROS_DEBUG("Calling Zivid::HDR::combineFrames with %zd frames", frames.size());
-      return Zivid::HDR::combineFrames(begin(frames), end(frames));
-    }
-    else
-    {
-      return frames[0];
-    }
-  }();
-  publishFrame(std::move(frame));
+  publishFrame(Zivid::HDR::capture(camera_, settings));
```
eskaur: Nice. This is pretty short and clean now :)

zivid_camera/src/zivid_camera.cpp
eskaur: Did you agree on whether or not to do this?

zivid_camera/srv/CaptureAssistantSuggestSettings.srv
runenordmo: Yes, so they can just disable/enable frames in dynamic reconfigure based on their preference.

zivid_samples/src/sample_capture_assistant.cpp
```diff
@@ -0,0 +1,74 @@
...
+    if (!cmd)                                                                                                          \
+    {                                                                                                                  \
+      throw std::runtime_error{ "\"" #cmd "\" failed!" };                                                              \
+    }                                                                                                                  \
+  } while (false)
+
+namespace
+{
+const ros::Duration default_wait_duration{ 30 };
+constexpr auto ca_suggest_settings_service_name = "/zivid_camera/capture_assistant/suggest_settings";
```
runenordmo: You set this duration and service name parameter in both the C++ and the Python sample. Perhaps you could fetch these parameters from the parameter server, using the launch file?

This way you could avoid syncing the parameters in all three places, and also avoid recompiling if you change the parameters.

zivid_camera/src/zivid_camera.cpp
apartridge: This was discussed below, in sample_capture_assistant.cpp.

zivid_camera/test/test_zivid_camera.cpp
```diff
@@ -36,11 +50,14 @@ class ZividNodeTest : public testing::Test
...
   static constexpr auto capture_service_name = "/zivid_camera/capture";
   static constexpr auto capture_2d_service_name = "/zivid_camera/capture_2d";
+  static constexpr auto capture_assistant_suggest_settings_service_name = "/zivid_camera/capture_assistant/"
+                                                                          "suggest_settings";
   static constexpr auto color_camera_info_topic_name = "/zivid_camera/color/camera_info";
   static constexpr auto color_image_color_topic_name = "/zivid_camera/color/image_color";
   static constexpr auto depth_camera_info_topic_name = "/zivid_camera/depth/camera_info";
   static constexpr auto depth_image_raw_topic_name = "/zivid_camera/depth/image_raw";
   static constexpr auto points_topic_name = "/zivid_camera/points";
+  static constexpr size_t num_dr_capture_servers = 10;
```
runenordmo: I think all the above parameters could be set from the launch file. This way we could avoid syncing both here and in the samples if we change the wait duration, topic names or the number of dynamic reconfigure services.

I think we could keep it like this if we are certain that these parameters will not change (the user can use `remap` if they need another topic name for instance for the points). I am also open to doing this as a separate issue, as it would require some other changes, for instance in the samples.

apartridge:
The CI issue directly related to the test seems fixed now. Let me know if you have any other comments. Especially feedback on the Getting started changes to the readme.

zivid_camera/src/zivid_camera.cpp
```diff
@@ -237,6 +208,8 @@ ZividCamera::ZividCamera(ros::NodeHandle& nh, ros::NodeHandle& priv)
   is_connected_service_ = nh_.advertiseService("is_connected", &ZividCamera::isConnectedServiceHandler, this);
   capture_service_ = nh_.advertiseService("capture", &ZividCamera::captureServiceHandler, this);
   capture_2d_service_ = nh_.advertiseService("capture_2d", &ZividCamera::capture2DServiceHandler, this);
+  capture_assistant_suggest_settings_service_ = nh_.advertiseService(
+      "capture_assistant/suggest_settings", &ZividCamera::captureAssistantSuggestSettingsServiceHandler, this);
```
runenordmo: These parameters are set both in the node, tests and the C++ and Python sample. I suggest setting the parameters in the launch file, such that we don't have to sync in all places.

If we know that these parameters will not change, if think we can keep it as is (if the user wants to change the topic names for instance, this can easily be done with `remap`). Alternatively, I am also open to do we do that in another issue, as it will require changes in the test, samples and launch files, and is not related to Capture Assistant functionality.

eskaur:
> The CI issue directly related to the test seems fixed now. Let me know if you have any other comments. Especially feedback on the Getting started changes to the readme.

"Getting started" page looks good now. Ran all the samples, and they work great.
Looked over the code as well (except the tests). No complaints from me. If Rune has also had a good look a the code, I think you can merge.

zivid_samples/src/sample_capture_assistant.cpp
apartridge: I think we want the samples simple and standalone, so the service name should be visible in the samples. Having too much code/constants that is shared and passed along makes the samples less clear. For example, user may just duplicate our sample cpp or py and get started. They probably dont want to use our sample.launch file.

apartridge:
Thanks for review. Squashed commits now, so need re-approval @eskaur and @runenordmo.

zivid_samples/src/sample_capture_assistant.cpp
runenordmo: I agree that simplicity of the samples can be worth the trade-off of having to update these constants several places in case they change.

PR #13, opened 2019-12-13
fix typo in package.xml

knorth55:
`website>` -> `website`

### DIFF
--- a/zivid_camera/package.xml
+++ b/zivid_camera/package.xml
@@ -5,7 +5,7 @@
   <description>Driver for using the Zivid 3D cameras in ROS.</description>
   <maintainer email="support@zivid.com">Zivid</maintainer>
   <license>BSD3</license>
-  <url type="website>">https://zivid.com</url>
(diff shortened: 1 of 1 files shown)


apartridge:
Hi @knorth55, thanks for this PR. Unfortunately, we do not currently have a Contributor License Agreement in place. This will hopefully be in place in the not too distant future. In the mean time, since this is a quick fix I have made an issue for this (https://github.com/zivid/zivid-ros/issues/22) and we will fix it very soon.

knorth55:
This package is BSD3 License, so I think we don't need that kind of Contributor license agreement, probably.
Anyway, it is just a typo fix, so you can commit by yourself.

https://github.com/zivid/zivid-ros/blob/master/LICENSE
https://github.com/zivid/zivid-ros/blob/master/zivid_camera/package.xml#L7

apartridge:
Fixed this in PR https://github.com/zivid/zivid-ros/pull/26. Thanks again for discovering this @knorth55.

Issue #31, opened 2020-06-29
zivid_camera_with_settings.launch some times does not set capture settings correctly


apartridge:
When running this sample launch file
https://github.com/zivid/zivid-ros/blob/master/zivid_samples/launch/zivid_camera_with_settings.launch

It can be observed that some times not all the settings are set. Normally, the "dynparam set_from_parameters" nodes updates the capture config as soon as the zivid_camera node is running, and then the node exits. However, some times the dynparam nodes hang forever, and never updates the config.

apartridge:
We have migrated to ROS2 so this issue isnt relevant anymore.

PR #32, opened 2020-07-08
Support sdk2

apartridge:
PR for updating ROS driver to support SDK 2.


zivid_camera/src/settings_generator.cpp
```diff
@@ -201,30 +281,28 @@ class DynamicReconfigureCfgGenerator
...
+    using ValueType = typename SettingsNode::ValueType;
 
-    const auto cfg_id = "cfg." + convertSettingsPathToConfigPath(s.path);
-    const auto setting_node_class_name = convertSettingsPathToZividClassName(zivid_settings_class_name_, s.path);
+    const auto cfg_id = "cfg." + convertSettingsPathToConfigPath<SettingsRootGroup, SettingsNode>();
+    const auto setting_node_class_name = fullyQualifiedZividTypeName<SettingsNode>();
     ss_ << "  s.set(" + setting_node_class_name + "{ ";
 
-    if constexpr (std::is_same_v<VT, std::size_t>)
+    if constexpr (std::is_same_v<ValueType, std::size_t>)
```
nedrebo, line 301: Nice

runenordmo:
As far as I can see, only the last commit in the PR is related to updating to sdk 2.0.

I guess there in the same PR because you wanted those clean-ups in first, without the overhead of creating another PR that needed to be merged first?

zivid_camera/src/settings_generator.cpp
```diff
@@ -49,6 +50,23 @@ struct DependentFalse : std::false_type
 
 void writeToFile(const std::string& file_name, const std::string& text)
 {
+  if (std::filesystem::exists(file_name))
+  {
+    std::ifstream file(file_name);
+    if (!file || !file.is_open())
```
runenordmo, line 49: Nice to avoid the unnecessary re-build here :+1: 
Curious; Is it needed to have both checks in `!file || !file.is_open()`?

apartridge: no it is not needed to have both checks, either one will work, i can change to just is_open.

apartridge:
> As far as I can see, only the last commit in the PR is related to updating to sdk 2.0.
> 
> I guess there in the same PR because you wanted those clean-ups in first, without the overhead of creating another PR that needed to be merged first?

Yes, this could have been split. To save time I think we can just review and merge them together.

zivid_camera/src/zivid_camera.cpp
```diff
@@ -518,11 +518,11 @@ sensor_msgs::PointCloud2ConstPtr ZividCamera::makePointCloud2(const std_msgs::He
...
-  msg->fields.push_back(createPointField("x", 0, 7, 1));
-  msg->fields.push_back(createPointField("y", 4, 7, 1));
-  msg->fields.push_back(createPointField("z", 8, 7, 1));
-  msg->fields.push_back(createPointField("c", 12, 7, 1));
-  msg->fields.push_back(createPointField("rgb", 16, 7, 1));
+  msg->fields.push_back(createPointField("x", 0, sensor_msgs::PointField::FLOAT32, 1));
+  msg->fields.push_back(createPointField("y", 4, sensor_msgs::PointField::FLOAT32, 1));
+  msg->fields.push_back(createPointField("z", 8, sensor_msgs::PointField::FLOAT32, 1));
+  msg->fields.push_back(createPointField("c", 12, sensor_msgs::PointField::FLOAT32, 1));
+  msg->fields.push_back(createPointField("rgb", 16, sensor_msgs::PointField::FLOAT32, 1));
```
runenordmo: Nice!
Why is the "rgb" field not `sensor_msgs::PointField::UINT8, 3`?

apartridge: This is consistent with how it was before, type number 7. I think it is for PCL compatibility reasons. I just changed to using the enum here instead of 7.

continuous-integration/test.sh
```diff
@@ -7,13 +7,13 @@ SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )" || exit $?
 cd ~/catkin_ws || exit $?
 
 echo "Installing Zivid API config file"
-install -D "$SCRIPT_DIR"/ZividAPIConfigCPU.yml "$HOME"/.config/Zivid/API/Config.yml || exit $?
+install -D "$SCRIPT_DIR"/ZividAPIConfigComputeDeviceAny.yml "$HOME"/.config/Zivid/API/Config.yml || exit $?
```
runenordmo, line 10: Can you comment (ideally also in the commit message) why the `ZividAPIConfigCPU` was replaced with `ZividAPIConfigComputeDeviceAny`?

zivid_camera/src/settings_generator.cpp
```diff
@@ -5,52 +5,45 @@
...
+
+template <typename T, typename... Ts>
+struct IsInTuple<T, std::tuple<Ts...>> : IsInList<T, Ts...>
 {
 };
 
 void writeToFile(const std::string& file_name, const std::string& text)
 {
-  if (std::filesystem::exists(file_name))
+  if (boost::filesystem::exists(file_name))
```
runenordmo, line 46: Why are you using `boost::filesystem` rather than `std::filesystem`? We require a C++17 compiler

zivid_camera/src/zivid_camera.cpp
runenordmo: Ok, I guess we can do something like [pcl::fromROSMsg(const sensor_msgs::PointCloud2 & cloud, pcl::PointCloud< T > &pcl_cloud)](http://docs.ros.org/kinetic/api/pcl_conversions/html/namespacepcl.html#af662c7d46db4cf6f7cfdc2aaf4439760) to confirm. 
Maybe you did already?

zivid_camera/src/settings_generator.cpp
apartridge: std::filesystem does not work on gcc-6 which otherwise should be c++17 compliant, due to this error
https://stackoverflow.com/questions/39231363/fatal-error-filesystem-no-such-file-or-directory
So I swapped to boost which we already use for other stuff to avoid this problem.

continuous-integration/test.sh
apartridge: It's not an important change. Its just to be even more flexible for the future. No point in saying either CPU or GPU if we in the future run these tests with a discrete GPU.

zivid_camera/src/settings_generator.cpp
runenordmo: Then [this commit ](https://github.com/zivid/zivid-ros/pull/32/commits/8c9a10fa21d91868a742847bb69931ae5d387527) wouldn't build on gcc-6, I guess?

apartridge: That is true, will fix

runenordmo:
image

Dynamic reconfigure looks nice :+1: 

What is the reason we do not have `processing` as a namespace in the same way as `acquisition`? 
- Is it because settings is the top level namespace, and that it could contain other things than `processing`? Is it a better match for the settings datamodel?

runenordmo:
Btw, in my [previous comment](https://github.com/zivid/zivid-ros/pull/32#issuecomment-671903274) the defaults look correct, but for instance the range for the brightness looks wrong (I captured with an Amber).
Is the range inteded to be also for the dynamic reconfigure or is it the largest range?

zivid_camera/include/zivid_camera.h
```diff
@@ -42,74 +43,59 @@ class ZividCamera
...
+  using Capture2DSettingsController =
+      CaptureSettingsController<Zivid::Settings2D, Settings2DConfig, Settings2DAcquisitionConfig>;
 
   ros::NodeHandle nh_;
   ros::NodeHandle priv_;
   ros::Timer camera_connection_keepalive_timer_;
   CameraStatus camera_status_;
-  std::unique_ptr<CaptureGeneralConfigDRServer> capture_general_config_dr_server_;
-  bool use_latched_publisher_for_points_;
+  bool use_latched_publisher_for_points_xyz_;
```
runenordmo, line 80: Is there a need to keep all these bools as member functions? I guess they're just needed in the constructor?

zivid_camera/include/capture_settings_controller.h
```diff
@@ -0,0 +1,61 @@
...
+/// <summary>Controller that manages dynamic_reconfigure nodes for Settings and Settings2D</summary>
+/// <remarks>
+/// This is a templated class that handles Zivid::Settings and Zivid::Settings2D configuration. Note that
+/// within the Zivid SDK Settings and Settings2D are similar in structure, but there are some differences
+/// in what kind of settings are provided.
+///
+/// This controller sets up the dynamic_reconfigure nodes "settings/", "settings/acquisition_0" and so
+/// on, handles callbacks and updates the internal config state. It also provides methods to convert
+/// from/to Zivid::Settings/Settings2D objects.
+/// </remarks>
```
runenordmo, line 29: Nice way of handling the per-acquisition settings and the other settings

zivid_camera/src/zivid_camera.cpp
```diff
@@ -449,40 +419,59 @@ bool ZividCamera::isConnectedServiceHandler(IsConnected::Request&, IsConnected::
...
+  if (publish_points_xyz || publish_points_xyzrgba || publish_color_img || publish_depth_img || publish_snr_img)
   {
     const auto header = makeHeader();
-    const auto point_cloud = frame.getPointCloud();
+    auto point_cloud = frame.pointCloud();
+
+    // Transform point cloud from millimeters (Zivid SDK) to meter (ROS).
+    const float scale = 0.001f;
+    const auto transformationMatrix = Zivid::Matrix4x4{ scale, 0, 0, 0, 0, scale, 0, 0, 0, 0, scale, 0, 0, 0, 0, 1 };
+    point_cloud.transform(transformationMatrix);
```
runenordmo, line 436: Nice, that we could use the transformation for the mm to m scaling :+1:

zivid_camera/src/zivid_camera.cpp
```diff
@@ -449,40 +419,59 @@ bool ZividCamera::isConnectedServiceHandler(IsConnected::Request&, IsConnected::
...
-
-    if (publish_color_img || publish_depth_img)
+    if (publish_points_xyzrgba)
     {
-      const auto camera_info = makeCameraInfo(header, point_cloud.width(), point_cloud.height(), camera_.intrinsics());
+      publishPointCloudXYZRGBA(header, point_cloud);
+    }
+    if (publish_color_img || publish_depth_img || publish_snr_img)
+    {
+      const auto intrinsics = Zivid::Experimental::Calibration::intrinsics(camera_);
```
runenordmo, line 448: Could the be any downsides to using experimental features in the wrapper?

zivid_camera/include/zivid_camera.h
apartridge: They are used other places, in fex shouldPublishPointsXYZRGBA, when we check if we should create and publish the different topics.

zivid_camera/src/zivid_camera.cpp
apartridge: This was moved into experimental in 2.x. We still need it in ROS. This probably means we cannot remove this function from Zivid::Experimental until we bump the major version again.

apartridge:
> image
> 
> Dynamic reconfigure looks nice 
> 
> What is the reason we do not have `processing` as a namespace in the same way as `acquisition`?
> 
> * Is it because settings is the top level namespace, and that it could contain other things than `processing`? Is it a better match for the settings datamodel?

Yes, I considered separate namespace for processing, even implemented it at one point, but I decided on doing it like this. Because it requires fewer changes if we add something other than processing/ on the top level. Now it will just work. Otherwise, we would need to add new code to make a dynamic_reconfigure server for that new top level node, which I think is not worth it. Also we would end up with one more server per new top-level type (possibly small performance hit).

apartridge:
> Btw, in my [previous comment](https://github.com/zivid/zivid-ros/pull/32#issuecomment-671903274) the defaults look correct, but for instance the range for the brightness looks wrong (I captured with an Amber, the range should be [0.0, 1.0]. Similar for exposure time).
> 
> The generated `SettingsAcquisitionConfigUtils.h` does look correct, though.
> 
> It looks like these values are the ones generated to be in `SettingsAcquisition.cfg`.

Yeah it could be that the ranges are not exposed correctly, I will look into it. It will fail when capturing if outside of the amber-range, but it is still an issue if the range is wrong.

zivid_camera/include/zivid_camera.h
runenordmo: Yes, I guess it's necessary since we then should publish even without subscribers. 
An alternative would be to always publish, but I think it's nicer the way we have it now; avoiding to do the work in f.ex. `publishPointCloudXYZRGBA()` unless subscribed to or latched.

zivid_samples/src/sample_capture.cpp
```diff
@@ -46,26 +46,35 @@ int main(int argc, char** argv)
...
+  auto points_sub = n.subscribe("/zivid_camera/points/xyzrgba", 1, on_points);
 
-  ROS_INFO("Enabling the reflection filter");
-  dynamic_reconfigure::Client<zivid_camera::CaptureGeneralConfig> capture_general_client("/zivid_camera/"
-                                                                                         "capture/general/");
-  zivid_camera::CaptureGeneralConfig config;
-  CHECK(capture_general_client.getDefaultConfiguration(config, default_wait_duration));
-  config.filters_reflection_enabled = true;
-  CHECK(capture_general_client.setConfiguration(config));
+  dynamic_reconfigure::Client<zivid_camera::SettingsConfig> settings_client("/zivid_camera/"
```
runenordmo, line 51: It seems a bit involved to use the sample. What if I only care about the pointcloud, and not about how the dynamic reconfigure setup works, and just want to `capture(Zivid::Settings settings)`?
(I guess this could be done similar to the `CaptureSettingsController<...>::setZividSettings(const ZividSettingsType& settings)`).

- I guess the reason the settings cannot be passed in this way is that it should also work for Python? And perhaps the interaction with the Capture Assistant's suggested settings? 

An alternative could perhaps be to have some sort of wrapper function in `zivid-sample`? 

HDR captures could be quite verbose here, but that might be something we'll have to accept. Could be interesting to have an [example which captures with settings from yml](https://github.com/zivid/zivid-cpp-samples/blob/master/source/Camera/Basic/CaptureWithSettingsFromYML/CaptureWithSettingsFromYML.cpp). But maybe we should consider that when handling #31.

zivid_camera/src/settings_generator.cpp
```diff
@@ -17,34 +17,27 @@
```
runenordmo, line 17: Maybe you can add a short decription at the top what this file/program does, that it auto-generates the .cfg files and needed header utilites to handle the different setting nodes? (and what it's used for)

zivid_camera/src/settings_generator.cpp
```diff
@@ -80,27 +73,91 @@ void writeToFile(const std::string& file_name, const std::string& text)
...
 class DynamicReconfigureCfgGenerator
 {
 public:
   DynamicReconfigureCfgGenerator(const std::string& class_name) : class_name_(class_name), insert_enabled_(false)
   {
   }
 
-  template <class ValueType>
+  template <typename SettingsNode>
+  auto cfgDefaultValue()
```
runenordmo, line 143: It's a bit confusing here that the `cfgDefaultValue()` actually gives the min values and false for bools.
  
Perhaps `cfgDummyDefaultValue()` or `cfgTemporaryDefaultValue()` could make it more clear? 
Or a comment in this line:
https://github.com/zivid/zivid-ros/blob/b050280eddb2e44715e5b581c3fddea4d3d233d3/zivid_camera/src/settings_generator.cpp#L235

zivid_camera/src/settings_generator.cpp
runenordmo: This is easier to spot if one looks at the `Generator` class, but that's 500 lines into the file

zivid_camera/src/settings_generator.cpp
```diff
@@ -205,7 +267,7 @@ class DynamicReconfigureCfgGenerator
     res << "gen = ParameterGenerator()\n";
     if (insert_enabled_)
     {
-      res << "gen.add(\"enabled\", bool_t, 0, \"When frame is enabled it will be included in captures\", "
+      res << "gen.add(\"enabled\", bool_t, 0, \"When this acquisition is enabled it will be included in captures\", "
              "False)\n";
     }
     res << ss_.str();
```
runenordmo, line 273: I know specifically this has not changed in the PR here, but is there any reason not to call `apply()` directly here (removing `ss_`), rather than calling apply in
https://github.com/zivid/zivid-ros/blob/b050280eddb2e44715e5b581c3fddea4d3d233d3/zivid_camera/src/settings_generator.cpp#L500
?
Same question for the header utils generator.

zivid_camera/src/settings_generator.cpp
```diff
@@ -453,65 +513,46 @@ class Generator
...
-  capture_general_gen.writeToFiles();
-  capture_frame_gen.writeToFiles();
-
-  const auto settings2D = Zivid::Settings2D{};
-  Generator capture2D_frame_gen(settings2D, "Capture2DFrame");
-  traverseSettings2DTree(settings2D, capture2D_frame_gen);
-  capture2D_frame_gen.insertEnabled();
-  capture2D_frame_gen.writeToFiles();
-
+  addSettingsType<Zivid::Settings>("Settings");
```
runenordmo, line 555: This looks very clean now :+1:

zivid_camera/src/zivid_camera.cpp
```diff
@@ -167,38 +190,31 @@ ZividCamera::ZividCamera(ros::NodeHandle& nh, ros::NodeHandle& priv)
...
-      std::make_unique<CaptureGeneralConfigDRServer>("capture/general", nh_, defaultSettings);
-
-  ROS_INFO("Setting up %d capture/frame_<n> dynamic_reconfigure servers", num_capture_frames);
-  for (int i = 0; i < num_capture_frames; i++)
-  {
-    capture_frame_config_dr_servers_.push_back(
-        std::make_unique<CaptureFrameConfigDRServer>("capture/frame_" + std::to_string(i), nh_, defaultSettings));
-  }
+  capture_settings_controller_ =
+      std::make_unique<Capture3DSettingsController>(nh_, camera_, "settings", max_capture_acquisitions);
```
runenordmo, line 203: Much tidier factoring out the settings controller related code from the `zivid_camera` :+1:

zivid_samples/src/sample_capture.cpp
apartridge: From the sample you must set the settings via dynamic_reconfigure (at least one acquisition). I agree that it becomes cumbersome, it is just the way it is now. There is no way to use Zivid::Settings or provide a YML from the clients code. We could maybe have a command that takes a YML file and applies it.

zivid_camera/src/settings_generator.cpp
apartridge: There is already a comment inside of cfgDefaultValue() that explains it, I don't think we should explain it again on the outside. What about renaming method to cfgFileDefaultValue() since this is literally the default value that is set in the .cfg file (which may be changed later by camera-specific default value).

zivid_camera/src/settings_generator.cpp
apartridge: I don't understand this comment fully. Apply is called from the outside for all of the settings, when we loop over all of the settings. This is why the output is stored in ss_. This str() is called later when writing the files to disk.

zivid_camera/include/zivid_camera.h
apartridge: Yeah it should be like it is now, since it is important for performance. We have a lot of different outputs (Z map, XYZ, XYZRGB etc) so the user can select what they want for their use case, without any performance hit (avoid copying unnecessary data and so on).

zivid_camera/test/test_zivid_camera.cpp
```diff
@@ -239,79 +246,204 @@ TEST_F(ZividNodeTest, testCapturePublishesTopics)
...
+
+  const auto pointCloud = captureViaSDKDefaultSettings();
+  const auto expectedSNR = pointCloud.copyData<Zivid::SNR>();
+  ASSERT_EQ(image->width, expectedSNR.width());
+  ASSERT_EQ(image->height, expectedSNR.height());
+  for (std::size_t i = 0; i < expectedSNR.size(); i++)
+  {
+    const auto expected = expectedSNR(i);
+    const float snr = *reinterpret_cast<float*>(image->data.data() + i * bytes_per_pixel);
+    ASSERT_EQ(snr, expected.value);
```
runenordmo, line 446: Why is there a epsilon for the checking the depth image, when it's not needed for the snr here?

zivid_camera/test/test_zivid_camera.cpp
```diff
@@ -239,79 +246,204 @@ TEST_F(ZividNodeTest, testCapturePublishesTopics)
...
+    const auto index = i * bytes_per_pixel;
+    ASSERT_EQ(image->data[index], expected.r);
+    ASSERT_EQ(image->data[index + 1], expected.g);
+    ASSERT_EQ(image->data[index + 2], expected.b);
+    ASSERT_EQ(image->data[index + 3], expected.a);
+    ASSERT_EQ(expected.a, 255);
+  }
+}
+
+TEST_F(CaptureOutputTest, testCaptureDepthImage)
```
runenordmo, line 394: Testing the snr and testing the depth looks very similar except the topic name and type T in `copyData<T>()`.
Perhaps we could re-use?

zivid_camera/test/test_zivid_camera.cpp
```diff
@@ -327,39 +459,41 @@ TEST_F(ZividNodeTest, testCaptureCameraInfo)
   auto depth_camera_info_sub =
       subscribe<sensor_msgs::CameraInfo>(depth_camera_info_topic_name, [&](const auto& r) { depth_camera_info = *r; });
 
-  enableFirst3DFrame();
-  zivid_camera::Capture capture;
-  ASSERT_TRUE(ros::service::call(capture_service_name, capture));
-  short_wait_duration.sleep();
+  std::optional<sensor_msgs::CameraInfo> snr_camera_info;
```
runenordmo, line 462: Looks like we have approx the same code three times in this test. Perhaps we can re-use in a function?

zivid_camera/test/test_zivid_camera.cpp
apartridge: For points we have converted the output from SDK (mm) to (m) so when we compare we divide SDK output by 1000 and compare it to ROS output.

zivid_camera/test/test_zivid_camera.cpp
```diff
@@ -397,18 +531,19 @@ TEST_F(ZividNodeTest, testCapture2D)
...
-      ASSERT_EQ(img.data[index], expectedRGB.r);
-      ASSERT_EQ(img.data[index + 1], expectedRGB.g);
-      ASSERT_EQ(img.data[index + 2], expectedRGB.b);
-      ASSERT_EQ(img.data[index + 3], expectedA);
-    };
-
-    for (const auto& expectedRGB : miscObjectsExpectedRGBs)
+    const auto frame2D = capture2DViaSDKDefaultSettings();
+    const auto expectedRGBA = frame2D.imageRGBA();
+    ASSERT_EQ(img.width, expectedRGBA.width());
```
runenordmo, line 536: Could consider re-using checking the rgba image which is equal in 2D and 3D.
If they're different types, we can just keep as is.

zivid_samples/scripts/sample_capture.py
```diff
@@ -16,23 +16,25 @@ def __init__(self):
 
         rospy.wait_for_service("/zivid_camera/capture", 30.0)
 
-        rospy.Subscriber("/zivid_camera/points", PointCloud2, self.on_points)
+        rospy.Subscriber("/zivid_camera/points/xyzrgba", PointCloud2, self.on_points)
```
runenordmo, line 19: Could be slightly annoying to someone who already uses this topic name in their code that we change the topic name here. But I think it makes sense, since the pointcloud format has actually change to not contain contrast (and since there are also updates like changing the iris to aperature).
Anyway the needed changes are apparent in the samples + readme.

zivid_camera/test/test_zivid_camera.cpp
runenordmo: Ahokey, so it's for the numerical error in that division. I'll resolve this now.

runenordmo:
Have looked through the entire PR now. Looks solid :+1: 
Will have a look when you have gone through the comments and added the fixup commits too.

What changes will be made going from supporting beta to official 2.0 release? I guess at least updating the readme and what version azure pipelines uses?

zivid_camera/src/settings_generator.cpp
runenordmo: Yes, let's go for `cfgFileDefaultValue()`.

zivid_camera/src/settings_generator.cpp
runenordmo: So it's because we want to traverse the settings tree outside and apply them to the generator? Which then combines this in the end with the boiler-plate part of the file

zivid_camera/test/test_zivid_camera.cpp
runenordmo: Closing this due to the other comment

> For points we have converted the output from SDK (mm) to (m) so when we compare we divide SDK output by 1000 and compare it to ROS output.

zivid_camera/test/test_zivid_camera.cpp
apartridge: Adding helper method for this

zivid_camera/src/zivid_camera.cpp
apartridge: I did the following to check and PCL also uses datatype 7 for rgb

  pcl::PointCloud<pcl::PointXYZRGB> cloud;
  sensor_msgs::PointCloud2 pc2;
  toROSMsg(cloud, pc2);

  ROS_INFO_STREAM("Stian debugging!");
  for (const auto& f : pc2.fields)
  {
    ROS_INFO_STREAM(f.name << " datatype=" << std::to_string(f.datatype) << " offset=" << std::to_string(f.offset));
  }

[ INFO] [1597404894.444920693]: x datatype=7 offset=0
[ INFO] [1597404894.444942774]: y datatype=7 offset=4
[ INFO] [1597404894.444961469]: z datatype=7 offset=8
[ INFO] [1597404894.444979463]: rgb datatype=7 offset=16

zivid_camera/test/test_zivid_camera.cpp
apartridge: Fixed

zivid_camera/src/settings_generator.cpp
apartridge: yes

apartridge:
> Have looked through the entire PR now. Looks solid 
> Will have a look when you have gone through the comments and added the fixup commits too.
> 
> What changes will be made going from supporting beta to official 2.0 release? I guess at least updating the readme and what version azure pipelines uses?

azure pipelines is already using the official 2.0 release. I don't expect any changes except changing the Changelog to remove beta.

zivid_samples/scripts/sample_capture.py
apartridge: Also we are adding one other "points" type "points/xyz" which does not have the color and is therefore more performant. So therefore I think it makes sense to rename the old one.

zivid_camera/src/zivid_camera.cpp
runenordmo: Nice :+1:

zivid_samples/scripts/sample_capture.py
runenordmo: I agree, resolving this one

zivid_samples/src/sample_capture.cpp
runenordmo: I added an issue for this in #33. You can add there if you have any thoughts.

runenordmo:
> > Have looked through the entire PR now. Looks solid
> > Will have a look when you have gone through the comments and added the fixup commits too.
> > What changes will be made going from supporting beta to official 2.0 release? I guess at least updating the readme and what version azure pipelines uses?
> 
> azure pipelines is already using the official 2.0 release. I don't expect any changes except changing the Changelog to remove beta.

Is there any reason the PR is still a draft?

zivid_camera/src/capture_settings_controller.cpp
```diff
@@ -12,28 +12,27 @@
...
     static_assert(std::is_same_v<ZividSettings, Zivid::Settings> || std::is_same_v<ZividSettings, Zivid::Settings2D> ||
                   std::is_same_v<ZividSettings, Zivid::Settings::Acquisition> ||
                   std::is_same_v<ZividSettings, Zivid::Settings2D::Acquisition>);
 
-    const auto config_min = zividSettingsToMinConfig<ConfigType>(defaultSettings);
+    const auto config_min = zividSettingsMinConfig<ConfigType, ZividSettings>(camera);
     dr_server_.setConfigMin(config_min);
 
-    const auto config_max = zividSettingsToMaxConfig<ConfigType>(defaultSettings);
+    const auto config_max = zividSettingsMaxConfig<ConfigType, ZividSettings>(camera);
```
runenordmo, line 29: Nice :+1: 
I am getting the correct default values with Zivid One now:
image

zivid_camera/test/test_zivid_camera.cpp
```diff
@@ -63,33 +65,45 @@ class ZividNodeTest : public testing::Test
...
-    template <class Type, class Fn>
-    static SubscriptionWrapper make(ros::NodeHandle& nh, const std::string& name, Fn&& fn)
+    static SubscriptionWrapper<Type> make(ros::NodeHandle& nh, const std::string& name)
     {
-      SubscriptionWrapper w;
-      boost::function<void(const boost::shared_ptr<const Type>&)> cb = [ptr = w.num_messages_.get(),
-                                                                        fn = std::move(fn)](const auto& v) mutable {
-        (*ptr)++;
-        fn(v);
+      auto w = SubscriptionWrapper<Type>();
```
runenordmo, line 74: Nice removal of the duplicated code :+1:

apartridge:
@runenordmo can you re-approve this PR?

runenordmo (approved):
Yes, re-approved :+1:

PR #36, opened 2020-10-05
Update to sdk 2.1

apartridge:


README.md
```diff
@@ -39,9 +39,9 @@ sudo apt-get install -y python-catkin-tools git
 
 ### Zivid SDK
 
-To use the ROS driver you need to download and install the "Zivid Core" package. SDK version 2.0 is supported.
-See [releases](https://github.com/zivid/zivid-ros/releases) for older ROS driver releases that supports
-older SDK versions.
+To use the ROS driver you need to download and install the "Zivid Core" package. Zivid SDK version 2.x is
```
runenordmo, line 42: Should we actually say 2.x here? For instance, this commit of the zivid-ros driver will likely work with 2.2, but not necessary support new features that are added.

For me this is okey, just wanted to bring it up.

README.md
```diff
@@ -342,8 +342,8 @@ cloud. For more information about HDR capture, visit our
 
 The capture settings available in the `zivid_camera` node matches the settings in the Zivid API.
 To become more familiar with the different settings and what they do, see the API reference for the
-[Settings](http://www.zivid.com/hubfs/softwarefiles/releases/2.0.0+7c4918cf-14/doc/cpp/classZivid_1_1Settings.html)
-and [Settings2D](http://www.zivid.com/hubfs/softwarefiles/releases/2.0.0+7c4918cf-14/doc/cpp/classZivid_1_1Settings2D.html)
+[Settings](http://www.zivid.com/hubfs/softwarefiles/releases/2.1.0+d2007e12-1/doc/cpp/classZivid_1_1Settings.html)
```
runenordmo, line 345: For DRY, it would be nice to make a variable for the version string.
Not sure if that's possible in Markdown. Do you know, @apartridge?

azure-pipelines.yml
```diff
@@ -29,13 +29,10 @@ stages:
...
-#        zivid_versions: ['']
-#        ros_distros: ['ros:melodic-ros-base-bionic', 'ros:kinetic-ros-base-xenial']
-#        compilers: ['g++-7']
+  - stage: Build_And_Test_Old_Zivid_Melodic_And_Kinetic
+    jobs:
+    - template: continuous-integration/azure-pipelines/templates/build_and_test.yml
+      parameters:
+        zivid_versions: ['2.0.0+7c4918cf-14']
+        ros_distros: ['ros:melodic-ros-base-bionic', 'ros:kinetic-ros-base-xenial']
+        compilers: ['g++-7']
```
runenordmo, line 38: What's the reason for testing only for one compiler on the older version(s)?
To save build time on the older version, also when there are more of them?

runenordmo (commented):
This looks quite good to me :+1: Added some comments/questions. 

Also wondering; Should we add some way of using `downsample()` (and perhaps `transform()`) from the ROS driver? I guess the most important thing is to emphasize that this works with 2.1, but might be something for us to consider.

README.md
apartridge: Yeah, I was thinking to maybe say 2.0 and 2.1 for now. I can do that. It is probable that we may have a breakage in 2.2 or later due to adding new setting types (enum) which is not handled by the ROS settings generator as of now.

README.md
apartridge: Unfortunately this is not possible

apartridge:
> This looks quite good to me  Added some comments/questions.
> 
> Also wondering; Should we add some way of using `downsample()` (and perhaps `transform()`) from the ROS driver? I guess the most important thing is to emphasize that this works with 2.1, but might be something for us to consider.

Yeah currently the only way to downsample or transform from ROS is to adjust the driver code. This is because transform and downsample is not exposed as settings, this is something we probably want in the future. Not just for ROS, but also for GeniCam.

azure-pipelines.yml
apartridge: To avoid combinatoric explosion I chose only oldest supported compiler for old SDK. Does not matter that much now since there is only 1 old version, but I think it is sufficient to test just the newest SDK for all the compilers.

README.md
runenordmo: Yes, I think also think say that we support 2.0 + 2.1.

README.md
runenordmo: Ah, unfortunate, that would be nice.
Then I'll just resolve this comment.

azure-pipelines.yml
runenordmo: Ok, makes sense :+1:

runenordmo:
> > This looks quite good to me  Added some comments/questions.
> > Also wondering; Should we add some way of using `downsample()` (and perhaps `transform()`) from the ROS driver? I guess the most important thing is to emphasize that this works with 2.1, but might be something for us to consider.
> 
> Yeah currently the only way to downsample or transform from ROS is to adjust the driver code. This is because transform and downsample is not exposed as settings, this is something we probably want in the future. Not just for ROS, but also for GeniCam.

You would prefer that we expose the options as settings before we update the ROS driver (+ GeniCam)? 
When we expose them as settings we will be able to do some internal optimizations as well, since we know that the user wants to downsample/transform before the capture happens.

runenordmo:
Should the `<version>2.0.0</version>` in [zivid_camera](https://github.com/zivid/zivid-ros/blob/b4aa18bc4da594d8427e7b8172612ec4238954e7/zivid_camera/package.xml#L4) and [zivid_samples](https://github.com/zivid/zivid-ros/blob/b4aa18bc4da594d8427e7b8172612ec4238954e7/zivid_samples/package.xml#L4) be updated to version `2.1.0`?

**Review done for this round, btw**

apartridge:
> > > This looks quite good to me  Added some comments/questions.
> > > Also wondering; Should we add some way of using `downsample()` (and perhaps `transform()`) from the ROS driver? I guess the most important thing is to emphasize that this works with 2.1, but might be something for us to consider.
> > 
> > 
> > Yeah currently the only way to downsample or transform from ROS is to adjust the driver code. This is because transform and downsample is not exposed as settings, this is something we probably want in the future. Not just for ROS, but also for GeniCam.
> 
> You would prefer that we expose the options as settings before we update the ROS driver (+ GeniCam)?
> When we expose them as settings we will be able to do some internal optimizations as well, since we know that the user wants to downsample/transform before the capture happens.

I added a bug to the internal tracker to add downsample to settings. We could choose to expose it in ROS dynamic_reconfigure without having it in settings, but I think that would be a mistake. It would be best to expose it as a setting, and then let ROS handle it automagically, consistent with all the other settings. If we have ROS users who want downsample before then, they would have to make a small adjustment to the driver code.

apartridge:
Fixed comments by force-push

runenordmo:
> > > > This looks quite good to me  Added some comments/questions.
> > > > Also wondering; Should we add some way of using `downsample()` (and perhaps `transform()`) from the ROS driver? I guess the most important thing is to emphasize that this works with 2.1, but might be something for us to consider.
> > > 
> > > 
> > > Yeah currently the only way to downsample or transform from ROS is to adjust the driver code. This is because transform and downsample is not exposed as settings, this is something we probably want in the future. Not just for ROS, but also for GeniCam.
> > 
> > 
> > You would prefer that we expose the options as settings before we update the ROS driver (+ GeniCam)?
> > When we expose them as settings we will be able to do some internal optimizations as well, since we know that the user wants to downsample/transform before the capture happens.
> 
> I added a bug to the internal tracker to add downsample to settings. We could choose to expose it in ROS dynamic_reconfigure without having it in settings, but I think that would be a mistake. It would be best to expose it as a setting, and then let ROS handle it automagically, consistent with all the other settings. If we have ROS users who want downsample before then, they would have to make a small adjustment to the driver code.

I think that sounds good, maximizing consistency.

runenordmo (approved):

runenordmo (approved):

PR #38, opened 2021-01-05
Support for sdk 2.2

apartridge:


runenordmo:
This is likely not introduced in this PR, but when I was running the sample I got reflection removal enabled by default in the dynamic reconfigure, which should likely not be the case:
image

zivid_camera/test/test_zivid_camera.test
```diff
@@ -2,5 +2,6 @@
     <node name="zivid_camera" pkg="zivid_camera" type="zivid_camera_node" ns="zivid_camera" output="screen">
         <param name="file_camera_path" type="str" value="/usr/share/Zivid/data/FileCameraZividOne.zfc" />
     </node>
-    <test test-name="zivid_camera_test" pkg="zivid_camera" type="zivid_camera_test" time-limit="300.0" />
+    <test test-name="zivid_camera_test" pkg="zivid_camera" type="zivid_camera_test" time-limit="300.0"
+        launch-prefix="bash -c 'rosservice call --wait /zivid_camera/zivid_camera/set_logger_level ros.zivid_camera debug; $0 $@' " />
```
runenordmo, line 6: Was hoping there was a nice way to set the verbosity level directly on the node in the launch file, but it does look like it exists.

What do you think about doing this instead of the launch-prefix?:
```
<node pkg="rosservice" type="rosservice" name="set_zivid_camera_logger_level" args="call --wait /zivid_camera/zivid_camera/set_logger_level 'ros.zivid_camera' 'debug'" />
```

zivid_camera/src/settings_generator.cpp
```diff
@@ -78,6 +85,12 @@ void writeToFile(const std::string& file_name, const std::string& text)
   }
 }
 
+std::string toUpperCaseFirst(std::string value)
+{
+  value[0] = std::toupper(value[0]);
```
runenordmo, line 90: Should perhaps check the string is not empty here

zivid_camera/src/settings_generator.cpp
```diff
@@ -159,6 +196,10 @@ class DynamicReconfigureCfgGenerator
     {
       return SettingsNode::validRange().min();
     }
+    if constexpr (IsEnumSetting<SettingsNode>::value)
+    {
+      return *SettingsNode::validValues().begin();
```
runenordmo, line 201: So validValues will here return a `std::set`, and it's implicit that the first element of the set is the default value (until the default value can be gotten from the camera)?

zivid_camera/src/settings_generator.cpp
```diff
@@ -136,6 +149,30 @@ std::string fullyQualifiedZividTypeName()
...
+  // zivid_camera::Settings_X. Since we may have several different enum settings with the
+  // same enum value X, we need to scope the enum name with the full path to the setting.
+  // This ensures there are no collisions now or in the future.
+  const auto value_str = toUpperCaseFirst(SettingsNode{ value }.toString());
+  const auto path = boost::replace_all_copy<std::string>(SettingsNode::path, "/", "");
+  return path + value_str;
+}
+
+template <typename SettingsNode>
+std::string generateEnumConstant(typename SettingsNode::ValueType value)
```
runenordmo, line 167: Nice, very easy to read :+1:

zivid_camera/test/test_zivid_camera.test
apartridge: i tried something like that first and could not get it to work. Have you tried this on your PC?

zivid_camera/src/settings_generator.cpp
```diff
@@ -240,6 +290,16 @@ class DynamicReconfigureCfgGenerator
     const auto type_name = rosTypeName<decltype(convertValueToRosValue(default_value))>();
     const auto default_value_str = valueTypeToRosTypeString(default_value);
 
+    if constexpr (IsEnumSetting<SettingsNode>::value)
+    {
+      const auto valid_values = node.validValues();
+      std::vector<std::string> enum_constants;
+      std::transform(valid_values.cbegin(), valid_values.cend(), std::back_inserter(enum_constants),
+                     [&](const auto value) { return generateEnumConstant<SettingsNode>(value); });
+      ss_ << rosGeneratedEnumVariableName(setting_name) << " = gen.enum([\n    "
```
runenordmo, line 299: I guess this `rosGeneratedEnumVariableName(setting_name)` has to be equal to the one you pass to `edit_emthod=` further down. Would be nice to save the name in a variable and use that both places, for DRY.
But I guess that can be difficult with the if constexpr's, i.e. we don't know that the setting is a enum?

zivid_camera/src/settings_generator.cpp
apartridge: Yes we gotta pick something, so we pick the first.

zivid_camera/src/settings_generator.cpp
```diff
@@ -248,6 +308,11 @@ class DynamicReconfigureCfgGenerator
       ss_ << ", " << valueTypeToRosTypeString(node.validRange().min()) << ", "
           << valueTypeToRosTypeString(node.validRange().max());
     }
+    else if constexpr (IsEnumSetting<SettingsNode>::value)
+    {
+      const auto max_index = node.validValues().size() - 1;
+      ss_ << ", 0, " << max_index << ", edit_method=" << rosGeneratedEnumVariableName(setting_name);
```
runenordmo, line 314: Could you you name the 0 in ", 0," explicitly? I guess it's the min value for the enum, but that might be difficult to know for someone reading this code.
Also I guess we are assuming here that the min enum value is 0, and not 3 for instance. I think naming the zero would make the assumption clearer.

runenordmo:
The `Stripe` engine option will now be visible in the dynamic reconfigure for Zivid One, although it's supported on Zivid One+.
I guess we are restricted by the way that we generate the enum setting in ROS, which will have to contain all possible enum values from the start (here two enum values).

zivid_camera/src/settings_generator.cpp
apartridge: Yeah, thats why I made it a function that is called from two places.

zivid_camera/test/test_zivid_camera.cpp
```diff
@@ -467,6 +467,36 @@ TEST_F(CaptureOutputTest, testCaptureSNRImage)
...
+  zivid_camera::SettingsConfig settings_cfg;
+  ASSERT_TRUE(settings_client.getDefaultConfiguration(settings_cfg, dr_get_max_wait_duration));
+  ASSERT_EQ(settings_cfg.experimental_engine, zivid_camera::Settings_ExperimentalEnginePhase);
+  settings_cfg.experimental_engine = zivid_camera::Settings_ExperimentalEngineStripe;
+  settings_cfg.processing_filters_reflection_removal_enabled = true;
+  settings_cfg.processing_filters_experimental_contrast_distortion_correction_enabled = true;
+  ASSERT_TRUE(settings_client.setConfiguration(settings_cfg));
+
+  zivid_camera::Capture capture;
+  // Capture fails here because file camera does not support Stripe engine
```
runenordmo, line 488: Yes, that makes it a bit difficult to write the test without a physical camera. We can add an issue for it in on the SDK side.

zivid_camera/src/settings_generator.cpp
apartridge: It is always non-empty as of now, but I can add a check for it.

zivid_camera/src/settings_generator.cpp
apartridge: Will fix

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 This project adheres to [Semantic Versioning](https://semver.org).
 
+## 2.2.0
+
+* Add support for version 2.2.0 of the Zivid SDK.
```
runenordmo, line 7: Should we add any info on the changes to the ROS driver here, or should we rely on the SDK 2.2 changelog and the updated settings in this README?

runenordmo:
Review done for this round, added some comments. Looks good :+1:

zivid_camera/src/settings_generator.cpp
runenordmo: Yes, looks good to me

zivid_camera/test/test_zivid_camera.test
apartridge: Actually it does work, the only problem is that since nodes are launched in parallell, we could be starting tests before the logger level is set. Which is the main reason that I did it like this, first enabling logging, then running the tests.

CHANGELOG.md
apartridge: I do not think we need to add info about the new stuff in 2.2. I was thinking maybe we should add that enum settings are now supported in ROS, but I thought it was covered by this point anyway.

zivid_camera/src/settings_generator.cpp
apartridge: fixed

zivid_camera/src/settings_generator.cpp
apartridge: done

zivid_camera/test/test_zivid_camera.test
runenordmo: Ah, I see. Let's keep it as is then, so we're sure it's the right order :+1:

CHANGELOG.md
runenordmo: Yes, I agree, it's apparent from the list of supported settings that enums are supported.

runenordmo (dismissed):

runenordmo (dismissed):

apartridge:
Added 1 commit that increases the test timeout (hopefully helps CI stability)

runenordmo:
> Added 1 commit that increases the test timeout (hopefully helps CI stability)

Any idea why that was needed (as opposed to before)?

apartridge:
> > Added 1 commit that increases the test timeout (hopefully helps CI stability)
> 
> Any idea why that was needed (as opposed to before)?

We added 1 test and it looks like sometimes the CI is slower and we hit the old limit, so we should bump it for now. To me it looks like the Azure CI machines are slow in general, in the CI it takes like 4-5 times as long as my local machine.

runenordmo:
> > > Added 1 commit that increases the test timeout (hopefully helps CI stability)
> > 
> > 
> > Any idea why that was needed (as opposed to before)?
> 
> We added 1 test and it looks like sometimes the CI is slower and we hit the old limit, so we should bump it for now. To me it looks like the Azure CI machines are slow in general, in the CI it takes like 4-5 times as long as my local machine.

I see. I'll re-approve now.

runenordmo (approved):

Issue #33, opened 2020-08-17
Add command for applying settings from YML file


runenordmo:

nedrebo:
Can you be more verbose? I am not sure what this is about.

runenordmo:
I think it's  quite involved providing settings for HDR in the ros sample.

We can add something in line with [the cpp sample capturing with samples from YML](https://github.com/zivid/zivid-cpp-samples/blob/master/source/Camera/Basic/CaptureWithSettingsFromYML/CaptureWithSettingsFromYML.cpp), or a command that let's you capture with settings from YML in ros.

Will probably require that #31 has been done first. This was part of a discussion in https://github.com/zivid/zivid-ros/pull/32#discussion_r469183151.

apartridge:
I agree that this should be improved. In SDK 2 we contain the whole HDR setting in one file, so we could add a command to load settings from a file, maybe also a launch parameter to initialize the settings.

PR #59, opened 2021-11-04
Add support for loading capture settings via YML file

danielcranston:
Hello, thanks for making this repository! Here's a PR adding some functionality I thought could be useful.

### Summary
This PR adds the ability to load and apply Zivid Settings via a `.yml` file, both at startup (via a rosparam) and at runtime (via a service).

This probably addresses https://github.com/zivid/zivid-ros/issues/33.

### Risks
1. No tests added for this functionality.
2. Requires a full path to the settings file.
3. Only tested on Zivid Two. I can't say for sure this works as intended on Zivid One since I don't have access to that hardware.
4. Currently, failure to load the `.yml` file at startup is not an error (will not throw), while failure when calling the service is. Maybe you have some preferences on this, please let me know :+1:

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,6 +198,13 @@ rosrun command:
 ROS_NAMESPACE=zivid_camera rosrun zivid_camera zivid_camera_node _frame_id:=camera1
 ```
 
+`settings_file_path` (string, default: "")
+> Specify the path to a settings `.yml` file. If specified, this file is used to construct the
+> [Zivid Settings](https://www.zivid.com/hubfs/softwarefiles/releases/2.5.0+19fa6891-1/doc/cpp/classZivid_1_1Settings.html#aa6fc7bf87c7a5f672b57397aab01ac3e)
+> and apply them at start up. All calls to the [capture](#capture) service will be made using these settings,
+> unless they are change via e.g. the [capture_assistant/suggest_settings](#capture_assistantsuggest_settings)
+> service.
+
 `file_camera_path` (string, default: "")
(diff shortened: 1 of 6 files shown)


apartridge:
Thanks @danielcranston. Zivid can take over this PR and complete the missing features, if that is OK for you? In addition to the things you mention, we should also add support for 2D-Settings.

Are you currently using the `rosparam` way of specifying Settings during startup of the node? I wonder if we should only support the service-call way, as it is a bit simpler, easier to test, and we avoid two ways of doing the same.

apartridge:
> Currently, failure to load the .yml file at startup is not an error (will not throw), while failure when calling the service is. Maybe you have some preferences on this, please let me know +1

It would be preferable that it did throw an error during startup if the provided .yml file was not loadable.

danielcranston:
Hi @apartridge,

> Zivid can take over this PR and complete the missing features, if that is OK for you?

Yeah that's fine with me, thanks for following up!

> Are you currently using the rosparam way of specifying Settings during startup of the node? I wonder if we should only support the service-call way, as it is a bit simpler, easier to test, and we avoid two ways of doing the same.

I am, but your argument for removing it makes sense so I'm not against it :+1:

apartridge:
Thanks @danielcranston  - I have started on a new branch for this here: https://github.com/zivid/zivid-ros/tree/add_support_for_loading_settings_from_file. I will hopefully make a PR for this early next week.

apartridge:
Made a separate PR here: https://github.com/zivid/zivid-ros/pull/60.
Feel free to review it and test it @danielcranston 
Closing this PR

PR #60, opened 2021-11-08
Add support for loading settings from file

apartridge:
Fixes issue #33.

Was based on PR https://github.com/zivid/zivid-ros/pull/59 from @danielcranston

### DIFF
--- a/zivid_camera/include/capture_settings_controller.h
+++ b/zivid_camera/include/capture_settings_controller.h
@@ -2,6 +2,7 @@
 
 #include "auto_generated_include_wrapper.h"
 
+#include <Zivid/CameraInfo.h>
 #include <Zivid/Settings.h>
 #include <Zivid/Settings2D.h>
 
@@ -50,6 +51,7 @@ private:
   using SettingsAcquisitionConfigTypeDRServer =
       ConfigDRServer<SettingsAcquisitionConfigType, typename ZividSettingsType::Acquisition>;
   std::string config_node_name_;
+  Zivid::CameraInfo m_cameraInfo;
   std::unique_ptr<SettingsConfigTypeDRServer> general_config_dr_server_;
   std::vector<std::unique_ptr<SettingsAcquisitionConfigTypeDRServer>> acquisition_config_dr_servers_;
 };
--- a/zivid_camera/include/capture_settings_controller.h
+++ b/zivid_camera/include/capture_settings_controller.h
@@ -51,7 +51,7 @@ private:
   using SettingsAcquisitionConfigTypeDRServer =
       ConfigDRServer<SettingsAcquisitionConfigType, typename ZividSettingsType::Acquisition>;
   std::string config_node_name_;
-  Zivid::CameraInfo m_cameraInfo;
+  Zivid::CameraInfo camera_info_;
   std::unique_ptr<SettingsConfigTypeDRServer> general_config_dr_server_;
   std::vector<std::unique_ptr<SettingsAcquisitionConfigTypeDRServer>> acquisition_config_dr_servers_;
 };
--- a/zivid_camera/src/capture_settings_controller.cpp
+++ b/zivid_camera/src/capture_settings_controller.cpp
@@ -98,7 +98,7 @@ CaptureSettingsController<ZividSettingsType, SettingsConfigType,
                                                                                     Zivid::Camera& camera,
                                                                                     const std::string& config_node_name,
                                                                                     std::size_t num_acquisition_servers)
-  : config_node_name_{ config_node_name }, m_cameraInfo{ camera.info() }
+  : config_node_name_{ config_node_name }, camera_info_{ camera.info() }
 {
   general_config_dr_server_ = std::make_unique<SettingsConfigTypeDRServer>(config_node_name_, nh, camera);
 
@@ -163,7 +163,7 @@ void CaptureSettingsController<ZividSettingsType, SettingsConfigType, SettingsAc
     throw std::runtime_error(error.str());
   }
 
-  const auto filledSettings = fillInUnsetWithCameraDefault(settings, m_cameraInfo);
+  const auto filledSettings = fillInUnsetWithCameraDefault(settings, camera_info_);
   ROS_DEBUG_STREAM("Updating settings to " << filledSettings);
   general_config_dr_server_->setConfig(zividSettingsToConfig<SettingsConfigType>(filledSettings));
 
--- a/zivid_camera/src/capture_settings_controller.cpp
+++ b/zivid_camera/src/capture_settings_controller.cpp
@@ -10,6 +10,33 @@
 
 #include <dynamic_reconfigure/server.h>
 
+namespace
+{
+template <typename Node>
+void recursivelyFillInUnsetWithCameraDefault(Node& node, const Zivid::CameraInfo& cameraInfo)
+{
+  if constexpr (Node::nodeType == Zivid::DataModel::NodeType::group ||
+                Node::nodeType == Zivid::DataModel::NodeType::leafDataModelList)
+  {
+    node.forEach([&cameraInfo](auto& child) { recursivelyFillInUnsetWithCameraDefault(child, cameraInfo); });
+  }
+  else if (!node.hasValue())
+  {
+    static_assert(Node::nodeType == Zivid::DataModel::NodeType::leafValue);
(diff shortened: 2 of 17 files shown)


danielcranston (commented):
Thanks for getting to this so quickly.

Getting rid of the rosparam for specifying a settings file at startup (and relying only on the service) sounds good to me, and that `recursivelyFillInUnsetWithCameraDefault` machinery really helps with user experience :+1: 

One potentially surprising behavior is that if you `load_settings_from_file`, specifying a file that omits some setting, that value will be force-set to the default value (due to `FillInUnsetWithCameraDefault`) _even if that setting has previously been set_.

Maybe a more intuitive behavior would be for settings that are not explicitly mentioned in the incoming setting file to be left at their current value, not be hard-set back to the default value? Essentially, `FillInUnsetWithCurrentValues`. I don't have any strong opinions here though, just thought I'd bring it up.

zivid_camera/include/capture_settings_controller.h
```diff
@@ -50,6 +51,7 @@ class CaptureSettingsController
   using SettingsAcquisitionConfigTypeDRServer =
       ConfigDRServer<SettingsAcquisitionConfigType, typename ZividSettingsType::Acquisition>;
   std::string config_node_name_;
+  Zivid::CameraInfo m_cameraInfo;
```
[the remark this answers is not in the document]
apartridge: Thanks, will fix

zivid_camera/srv/LoadSettingsFromFile.srv
```diff
@@ -0,0 +1,2 @@
+string file_path
+---
```
runenordmo, line 2: Would it not make sense to return a bool here indicating whether loading the settings file was successful or not?

zivid_camera/src/zivid_camera.cpp
```diff
@@ -403,7 +407,23 @@ bool ZividCamera::captureAssistantSuggestSettingsServiceHandler(CaptureAssistant
...
   return true;
-}  // namespace zivid_camera
+}
+
+bool ZividCamera::loadSettingsFromFileServiceHandler(LoadSettingsFromFile::Request& req,
+                                                     LoadSettingsFromFile::Response&)
+{
+  ROS_DEBUG_STREAM(__func__ << ": Request: " << req);
+  capture_settings_controller_->setZividSettings(Zivid::Settings{ req.file_path.c_str() });
+  return true;
```
runenordmo, line 417: I guess since you are throwing inside `setZividSettings()` if setting the settings does not complete, you cannot answer a `false` bool on the load setting service.

zivid_camera/src/capture_settings_controller.cpp
```diff
@@ -10,6 +10,33 @@
...
+  {
+    static_assert(Node::nodeType == Zivid::DataModel::NodeType::leafValue);
+    node = Zivid::Experimental::SettingsInfo::defaultValue<Node>(cameraInfo);
+  }
+}
+
+template <typename ZividSettingsType>
+ZividSettingsType fillInUnsetWithCameraDefault(const ZividSettingsType& settings, const Zivid::CameraInfo& cameraInfo)
+{
+  ZividSettingsType copy{ settings };
```
runenordmo, line 33: Why not pass `settings` by-value if you anyway making a copy?

zivid_camera/src/zivid_camera.cpp
apartridge: Exceptions are used for errors, also other places like for `capture`. The node does not shut down if a service function call throws, it is caught by something in ROS which makes the service call error out:

```
$ rosservice call /zivid_camera/load_settings_from_file "file_path: 'x'" 
ERROR: service [/zivid_camera/load_settings_from_file] responded with an error: b'Failed loading Settings from "x"'
$ rosservice call /zivid_camera/load_settings_from_file "file_path: 'y'" 
ERROR: service [/zivid_camera/load_settings_from_file] responded with an error: b'Failed loading Settings from "y"'
$ rosservice call /zivid_camera/load_settings_from_file "file_path: 'z'" 
ERROR: service [/zivid_camera/load_settings_from_file] responded with an error: b'Failed loading Settings from "z"'
```

```
[INFO] [1636467401.453709]: Calling capture service
[ INFO] [1636467401.457850638]: Capturing with 3 acquisition(s)
[INFO] [1636467403.330188]: PointCloud received
[INFO] [1636467403.344540]: 2D color image received
[ERROR] [1636468166.470776730]: Exception thrown while processing service call: Failed loading Settings from "x"
[ERROR] [1636468172.394322735]: Exception thrown while processing service call: Failed loading Settings from "y"
[ERROR] [1636468174.872632798]: Exception thrown while processing service call: Failed loading Settings from "z"
```

zivid_camera/src/capture_settings_controller.cpp
apartridge: Yeah could do that, but I think it doesn't matter much here, since at the call site I would anyway make the copy. Currently there is 1 copy, line 33. Passing by value there would be 1 copy and 1 move.
 I could also take settings by reference and modify it, but found returning a copy cleaner.

zivid_camera/srv/LoadSettingsFromFile.srv
apartridge: Answered below

zivid_camera/src/capture_settings_controller.cpp
```diff
@@ -10,6 +10,33 @@
...
+  if constexpr (Node::nodeType == Zivid::DataModel::NodeType::group ||
+                Node::nodeType == Zivid::DataModel::NodeType::leafDataModelList)
+  {
+    node.forEach([&cameraInfo](auto& child) { recursivelyFillInUnsetWithCameraDefault(child, cameraInfo); });
+  }
+  else if (!node.hasValue())
+  {
+    static_assert(Node::nodeType == Zivid::DataModel::NodeType::leafValue);
+    node = Zivid::Experimental::SettingsInfo::defaultValue<Node>(cameraInfo);
+  }
```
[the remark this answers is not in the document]
apartridge: Yes, previously we only called on setZividSettings() with output from CaptureAssistant::suggestSettings which always have all the nodes set.

zivid_camera/test/test_zivid_camera.cpp
```diff
@@ -816,6 +909,107 @@ class ZividCATest : public TestWithFileCamera
...
+    ASSERT_TRUE(ros::service::waitForService(serviceName, short_wait_duration));
+    CmdType cmd;
+    cmd.request.file_path = "/tmp/foo/bar";
+    ASSERT_FALSE(boost::filesystem::exists(cmd.request.file_path));
+    ASSERT_FALSE(ros::service::call(serviceName, cmd));
+
+    cmd.request.file_path = testDataDir() + "/settings/invalid_file.yml";
+    ASSERT_TRUE(boost::filesystem::exists(cmd.request.file_path));
+    ASSERT_FALSE(ros::service::call(serviceName, cmd));
+  }
```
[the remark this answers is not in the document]
apartridge: Yeah can split it, it gets a bit messy below though, because both functions then have to be called for Settings and Settings2D.

runenordmo:
No more comments from me, looks good :+1:

zivid_camera/src/zivid_camera.cpp
runenordmo: Ok, that looks reasonable :+1:

zivid_camera/test/test_zivid_camera.cpp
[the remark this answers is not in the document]
runenordmo: Ok, not so important, if it gets messy you can keep it as-is

runenordmo (approved):

zivid_camera/test/test_zivid_camera.cpp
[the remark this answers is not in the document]
apartridge: I split up this into two tests

apartridge:
The review got dismissed so you need to re-approve it @runenordmo

zivid_camera/test/test_zivid_camera.cpp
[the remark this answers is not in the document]
runenordmo: Nice! :+1:

runenordmo (approved):
Test became a bit more expressive, that's nice.
Still LGTM, good job! :+1:

Issue #34, opened 2020-08-19
Add launch file with all launch parameters


apartridge:
Request from user:

> I also believe that you should extend the launch file example to contain all the parameters available, so that end users do not have to read all the code to look for hidden functionality. A launch file with all the parameters available provides very concise documentation that lets the user know what it can change / configure.

PR #61, opened 2021-11-08
Add launch parameter to disable auto firmware update

apartridge:


README.md
```diff
@@ -191,13 +191,23 @@ ROS_NAMESPACE=zivid_camera rosrun nodelet nodelet standalone zivid_camera/nodele
...
 ```
 
+Or, if using `roslaunch` specify the parameter using `<param>`:
+
+```xml
+<launch>
+    <node name="zivid_camera" pkg="zivid_camera" type="zivid_camera_node" ns="zivid_camera" output="screen">
+        <param type="string" name="frame_id" value="camera_1" />
+    </node>
+</launch>
```
runenordmo, line 208: Nice to add the example of launching the zivid_camera node with parameters.

I think it would be nice to add the available launch parameters to the sample.launch file also, since it's most convenient to run the zivid_camera node (and the application code) from launch files. It would be nice to include the available options to the sample.launch file then.
https://github.com/zivid/zivid-ros/blob/1371fe90455bb8aa45c556ef21a3594abb22ad20/zivid_samples/launch/sample.launch#L4

runenordmo: Then we can add most/all available options, like `serial_number` and `update_firmware_automatically`, to that sample launch file.

runenordmo: Alternatively add a new zivid_camera launch file under https://github.com/zivid/zivid-ros/tree/master/zivid_camera/launch, and <include/> that launch file in the sample launch file.

runenordmo:
Looks good, added only one comment/suggestion :+1:

README.md
apartridge: Yes, might be a good idea to add a launch file with all the parameters. Here I just wanted to add an example of how to set it for a .launch file yourself. The disadvantage of having all of them in the launch file is that we duplicate the default value, I suppose. We already have an existing request for this: https://github.com/zivid/zivid-ros/issues/34, so I think we should not solve it now as part of this PR.

runenordmo: Ah, sorry didn't go through the current list of issues. #34 explains very well what I meant.
Sure, it's fine for me not to update it as part of this PR.

If it's fine by you, I could make a follow-up PR for #34. I missed this when I was using the driver for internal usage.

runenordmo (approved):

README.md
apartridge: Feel free to make a PR for this @runenordmo

PR #66, opened 2022-04-20
Add sample which loads settings from file

### LINKED ISSUE #65: Add sample to show how to load settings from yml file
We should have a small sample to show how to use the LoadSettingsFromFile.srv

torbsorb:
We want to show how you can use the
LoadSettingsFromFile service to set capture settings from
a yml file.

Closes #65

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,6 +291,8 @@ visible via dynamic_reconfigure, see section [Configuration](#configuration).
 When this service has returned you can invoke the [capture](#capture) service to trigger a 3D
 capture using these settings.
 
+See [Sample Capture with Settings from File](#sample-capture-with-settings-from-file) for code example.
+
 ### load_settings_2d_from_file
 [zivid_camera/LoadSettings2DFromFile.srv](./zivid_camera/srv/LoadSettings2DFromFile.srv)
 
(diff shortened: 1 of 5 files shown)


zivid_samples/scripts/sample_capture_with_settings_from_yml.py
```diff
@@ -0,0 +1,33 @@
...
+        rospy.init_node("sample_capture_with_settings_from_yml_py", anonymous=True)
+
+        rospy.loginfo("Starting sample_capture_with_settings_from_yml.py")
+
+        rospy.wait_for_service("/zivid_camera/capture", 30.0)
+
+        self.capture_service = rospy.ServiceProxy("/zivid_camera/capture", Capture)
+
+        rospy.loginfo("Loading settings from file")
+        self.load_settings_from_file = rospy.ServiceProxy("/zivid_camera/load_settings_from_file", LoadSettingsFromFile)
```
apartridge, line 22: Can you add `_service` to this var name to be consistent with other services?

README.md
```diff
@@ -593,6 +595,25 @@ rosrun zivid_samples sample_capture_2d_cpp
 rosrun zivid_samples sample_capture_2d.py
 ```
 
+### Sample Capture with Settings from File
+
+This sample shows how to [configure](#configuration) the 3D capture settings
+by invoking the [load_settings_from_file](#load_settings_from_file) service and
+how to invoke the [capture](#capture) service.
```
apartridge, line 602: Maybe you can also add here that the same idea also works for 2D settings, and  describe the difference fex that you must use `load_settings_2d_from_file` service? It is also possible to have it as a separate sample or even set both 2D and 3D in the same sample (with some comment to remove it if you don't want to set 2D settings).

torbsorb: I can write something about it. In the long run we want to achieve sample parity with the other languages. As such we also want to include it in the [capture tutorial](https://support.zivid.com/latest/academy/camera/capture-tutorial.html#load). It's unfortunate that the settings in ROS seem to be a state, as opposed to in Python, C++ and C#.

zivid_samples/src/sample_capture_with_settings_from_yml.cpp
```diff
@@ -0,0 +1,59 @@
...
+  ROS_INFO("Calling capture service");
+  zivid_camera::Capture capture;
+  CHECK(ros::service::call("/zivid_camera/capture", capture));
+}
+
+}  // namespace
+
+int main(int argc, char** argv)
+{
+  ros::init(argc, argv, "sample_capture_cpp");
```
apartridge, line 35: can you update this to be correct for this sample?

torbsorb: done

apartridge (approved):

PR #80, opened 2023-01-09
Fix broken CI status badge

apartridge:
There was a change in shields.io API, see issue
https://github.com/badges/shields/issues/8671. This change makes the build badge work again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -704,7 +704,7 @@ This FTP (Focused Technical Project) has received funding from the European Unio
 Horizon 2020 research and innovation programme under the project ROSIN with the
 grant agreement No 732287. For more information, visit [rosin-project.eu](http://rosin-project.eu/).
 
-[ci-badge]: https://img.shields.io/github/workflow/status/zivid/zivid-ros/ROS%20Commit/master
+[ci-badge]: https://img.shields.io/github/actions/workflow/status/zivid/zivid-ros/ROS-commit.yml?branch=master
 [ci-url]: https://github.com/zivid/zivid-ros/actions?query=workflow%3A%22ROS+Commit%22+branch%3Amaster+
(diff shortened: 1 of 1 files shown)


knatten:
Seems to work fine!

I'm not very experienced with this, but does anyone know why we use shields.io in the first place? On one of my side projects, I use GitHub directly like this:

```
![CI](https://github.com/knatten/DumbLang/actions/workflows/ci.yml/badge.svg)
```

It renders like this:

Screenshot 2023-01-09 at 17 39 42


This is documented at https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/adding-a-workflow-status-badge.

knatten (approved):

apartridge:
Maybe @trym-b remembers if there is a good reason to use shields.io? Else we can swap it out with the official one.

knatten:
Could be that GitHub didn't support this when we started using it or something.

trym-b:
> Maybe @trym-b remembers if there is a good reason to use shields.io? Else we can swap it out with the official one.

I think we should use the official ones. I do not recall why we exactly ended up selecting shields.io instead of something else.

nedrebo:
> > Maybe @trym-b remembers if there is a good reason to use shields.io? Else we can swap it out with the official one.
> 
> I think we should use the official ones. I do not recall why we exactly ended up selecting shields.io instead of something else.

Because if you have multiple badges from various vendors like we do in Python, then it looks messy and inconsistent unless you use a proxy provider that gives a uniform look.

nedrebo:
Seems we need the same fix in `zivid-python`.

nedrebo (approved):

nedrebo:
> Seems we need the same fix in `zivid-python`.

Any takers @eskaur :D

PR #85, opened 2023-06-22
Fix wrong port in README.md

iosb-ina-mr:
In the readme it says that the camera has to be connected to a USB3 port but the camera supports ethernet only by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ catkin build
 
 ## Getting started
 
-Connect the Zivid camera to your ethernet port on your PC. You can use the ZividListCameras command-line
+Connect the Zivid camera to your ethernet or USB3 port on your PC. You can use the ZividListCameras command-line
 tool available in the "Zivid Tools" package to confirm that your system has been configured correctly, and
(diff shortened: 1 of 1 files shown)


apartridge:
Hi @iosb-ina-mr, thanks for your PR. The Zivid One+ cameras use USB3, while the Zivid Two and Two+ cameras use Ethernet. So this text was indeed incorrect, we forgot to update this part after our release of the Zivid Two camera. Maybe you can change your commit to "ethernet or USB3 port"?

iosb-ina-mr:
> Hi @iosb-ina-mr, thanks for your PR. The Zivid One+ cameras use USB3, while the Zivid Two and Two+ cameras use Ethernet. So this text was indeed incorrect, we forgot to update this part after our release of the Zivid Two camera. Maybe you can change your commit to "ethernet or USB3 port"?

For sure, I was not aware of that. I have adjusted the PR.

apartridge (approved):

apartridge:
@iosb-ina-mr  I did a rebase of the PR in order to re-trigger the tests since they failed due to a flaky test.

PR #86, opened 2023-07-03
Remove ubuntu1804 support

apartridge:


continuous-integration/setup/setup_build_and_test.sh
```diff
@@ -28,7 +28,9 @@ fi
 
 function install_opencl_cpu_runtime {
 
-    # Install driver prerequisites
+    echo "Begin install of OpenCL CPU runtime"
```
eskaur, line 31: Btw I recently update `zivid-python` to a more modern CPU runtime, so that we no longer depend on a Dropbox file:
https://github.com/zivid/zivid-python/blob/04c0b59002e2eb7295fb4320788f29cb0192dc98/continuous-integration/linux/platform-dependent/common.sh#L5

You could consider doing that here as well (not in this PR).

apartridge: Yes, planning to do that also. I have been seeing some OpenCL CPU driver install issues in the ROS CI which is what motivated me to remove U18 testing.

PR #94, opened 2023-12-15
Update to sdk 2.11.0

johningve:


README.md
apartridge, line 1: Should also update the top of the readme that refers to versions supported. Search for 
"To use the ROS driver you need"

apartridge (approved):

PR #129, opened 2025-03-14
Misc 2025 03 07 hand eye calibration services

micragz:


README.md
```diff
@@ -277,6 +278,72 @@ service call. If the camera is not in `Connected` state the driver will attempt
...
+
+### infield_correction/reset
+[std_srvs/srv/Trigger](https://docs.ros2.org/latest/api/std_srvs/srv/Trigger.html)
+
+Resets the infield correction on the camera to factory settings.
+
+### infield_correction/start
+[std_srvs/srv/Trigger](https://docs.ros2.org/latest/api/std_srvs/srv/Trigger.html)
+
+Prepares for infield correction, clears any infield correction captures gathered so far in the `zivid_camera` node.
```
micragz, line 346: Mention calling this every time we want to start over again.

zivid_interfaces/srv/InfieldCorrectionCompute.srv
```diff
@@ -0,0 +1,33 @@
+---
+bool success
+string message
+
+# Returns true if infield correction has been started on the target node. It needs to be started before capturing and computing.
```
micragz, line 5: Refer to knowledge base

zivid_interfaces/msg/FixedPlacementOfCalibrationObjects.msg
```diff
@@ -0,0 +1,11 @@
+uint8 NONE = 0
+uint8 CALIBRATION_BOARD = 1
+uint8 FIDUCIAL_MARKERS = 2
+
+# NONE (default) or CALIBRATION_BOARD or FIDUCIAL_MARKER
+uint8 type
+
+FixedPlacementOfCalibrationBoard calibration_board
+
+string marker_dictionary
```
micragz, line 10: Default dictionary or not?

zivid_interfaces/srv/HandEyeCalibrationCalibrate.srv
```diff
@@ -0,0 +1,18 @@
...
+
+# Optionally, add fixed objects for low degrees-of-freedom (DOF) hand-eye calibration. Leave default for normal (6-DOF)
+# calibration.
+FixedPlacementOfCalibrationObjects fixed_objects
+---
+bool success
+string message
+
+geometry_msgs/Transform transform
+HandEyeCalibrationResidual[] residuals
```
micragz, line 18: Comment on order of residuals.

zivid_interfaces/srv/HandEyeCalibrationCalibrate.srv
```diff
@@ -0,0 +1,18 @@
+uint8 EYE_TO_HAND = 1
+uint8 EYE_IN_HAND = 2
+
+# EYE_TO_HAND or EYE_IN_HAND
+uint8 configuration
+
+# Indices of hand-eye captures to use in this hand-eye calibration. Leave empty to use all captures.
+int32[] capture_indices
```
micragz, line 8: capture id or capture handle?
Pose pair ID?

To avoid confusing different indices (SDK vs this capture list)

zivid_samples/src/sample_hand_eye_calibration.cpp
```diff
@@ -0,0 +1,346 @@
...
+    }
+    RCLCPP_INFO(
+      node->get_logger(), "Waiting for the %s service to appear...", client->get_service_name());
+  }
+  return client;
+}
+
+void set_settings(const std::shared_ptr<rclcpp::Node> & node)
+{
+  RCLCPP_INFO(node->get_logger(), "Setting parameter 'settings_yaml'");
```
micragz, line 99: Comment on proper settings

zivid_camera/include/zivid_camera/zivid_camera.hpp
```diff
@@ -169,6 +228,29 @@ class ZividCamera : public rclcpp::Node
     capture_assistant_suggest_settings_service_;
   rclcpp::Service<zivid_interfaces::srv::IsConnected>::SharedPtr is_connected_service_;
 
+  std::unique_ptr<InfieldCorrectionState> infield_correction_state_;
```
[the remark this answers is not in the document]
micragz: The new features have been refactored into separate files now.

zivid_interfaces/msg/FixedPlacementOfCalibrationObjects.msg
micragz: Removed and written as comment instead.

zivid_interfaces/srv/HandEyeCalibrationCalibrate.srv
micragz: Renamed to capture handle.

johningve (commented):
Started reviewing infield. Some comments

README.md
```diff
@@ -277,6 +278,123 @@ service call. If the camera is not in `Connected` state the driver will attempt
...
+### infield_correction/verify
+[zivid_interfaces/srv/InfieldCorrectionVerify.srv](./zivid_interfaces/srv/InfieldCorrectionVerify.srv)
+
+Verifies the current camera trueness based on a single capture. This service allows quick assessment of the quality of
+the infield correction on a camera, or the need for one if none exists already.
+
+Returns an indication of the dimension trueness at the location where the input data was captured. If the returned
+assessment indicates a trueness error that is above the threshold for your application, consider using
+[infield_correction/compute_and_write](#infield_correctioncompute_and_write) and related services to update the
+correction for the camera.
```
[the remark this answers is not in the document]
micragz: I tried to be consistent with our CLI tools and samples, and they take a capture, e.g.: https://github.com/zivid/zivid-python-samples/blob/master/source/camera/maintenance/verify_camera_in_field.py

Yes, this will work without start. I will try to clarify this a bit in the readme.

zivid_interfaces/srv/InfieldCorrectionVerify.srv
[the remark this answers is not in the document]
micragz: We have the `success` return code already. I don't think the other `status` codes are that useful programmatically, they're mainly used to get a readable message? But I can add them if you think they are valuable.

README.md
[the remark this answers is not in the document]
micragz: Updated readme here: 77f1f9553d2a09ee34f4f83313ea960fb07040f0

README.md
```diff
@@ -277,6 +278,123 @@ service call. If the camera is not in `Connected` state the driver will attempt
...
+### hand_eye_calibration/load
+[zivid_interfaces/srv/HandEyeCalibrationLoad.srv](./zivid_interfaces/srv/HandEyeCalibrationLoad.srv)
+
+Loads a working directory from a previous hand-eye calibration session. The captures and robot poses are loaded from the
+provided directory. The directory is opened as read-only, and no new captures can be made during this session. However,
+calls to the [hand_eye_calibration/calibrate](#hand_eye_calibrationcalibrate) service can be made to compute the hand-eye
+transform from the loaded data.
+
+### infield_correction/read
+[std_srvs/srv/Trigger](https://docs.ros2.org/latest/api/std_srvs/srv/Trigger.html)
```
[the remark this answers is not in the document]
micragz: Added time point here: 77f1f9553d2a09ee34f4f83313ea960fb07040f0

zivid_interfaces/srv/InfieldCorrectionVerify.srv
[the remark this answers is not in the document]
johningve: I'm not sure how many people would use them, but I imagine it would be annoying to find them missing in ROS if someone did want to use them.

README.md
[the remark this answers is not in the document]
johningve: I guess this README section needs updating as well to show the `zivid_interfaces/InfieldCorrectionRead`?

README.md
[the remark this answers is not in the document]
johningve: That is helpful. I think it is also good to be explicit here in the README that it will capture and *then* verify.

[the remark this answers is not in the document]
micragz: Updated readme: fcf92f9f8ba49be65e5f175559026f0d46dcc0cf

zivid_interfaces/srv/InfieldCorrectionVerify.srv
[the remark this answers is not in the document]
micragz: Yeah, that's fair enough, I added status here: fcf92f9f8ba49be65e5f175559026f0d46dcc0cf

micragz:
Added tests for infield correction.

micragz:
Planned to-do:

- Add Python samples
- Add hand-eye calibration tests
- Remove support for SDK older than 2.15.0.

apartridge:
> * Remove support for SDK older than 2.15.0.

If this is for hand-eye, then we should consider to make sepearate PR for infield and merge that first since then those on <2.15 can still get infield supported. But I think we can review both features together in this PR anyway.

README.md
```diff
@@ -277,6 +278,127 @@ service call. If the camera is not in `Connected` state the driver will attempt
...
+assessment indicates a trueness error that is above the threshold for your application, consider using
+[infield_correction/compute_and_write](#infield_correctioncompute_and_write) and related services to update the
+correction for the camera.
+
+### infield_correction/reset
+[std_srvs/srv/Trigger](https://docs.ros2.org/latest/api/std_srvs/srv/Trigger.html)
+
+Resets the infield correction on the camera to factory settings.
+
+### infield_correction/start
```
apartridge, line 346: Maybe list /start service first of the infield services? (To match what we do in handeye).

README.md
```diff
@@ -277,6 +278,127 @@ service call. If the camera is not in `Connected` state the driver will attempt
...
+[zivid_interfaces/srv/InfieldCorrectionVerify.srv](./zivid_interfaces/srv/InfieldCorrectionVerify.srv)
+
+Verifies the current camera trueness based on a single capture. The service first performs a capture with the camera,
+and then performs verification on that capture. This service allows quick assessment of the quality of the infield
+correction on a camera, or the need for one if none exists already.
+
+Returns an indication of the dimension trueness at the location where the input data was captured. If the returned
+assessment indicates a trueness error that is above the threshold for your application, consider using
+[infield_correction/compute_and_write](#infield_correctioncompute_and_write) and related services to update the
+correction for the camera.
```
apartridge, line 339: I wonder  why we have this one? Will it not be same as start/ + capture/ + compute/? Consider to remove it to simplify things.

README.md
```diff
@@ -277,6 +278,127 @@ service call. If the camera is not in `Connected` state the driver will attempt
...
+Resets the infield correction on the camera to factory settings.
+
+### infield_correction/start
+[std_srvs/srv/Trigger](https://docs.ros2.org/latest/api/std_srvs/srv/Trigger.html)
+
+Prepares for infield correction, and clears any infield correction captures gathered so far in the `zivid_camera` node.
+
+This service must be called before gathering captures using the [infield_correction/capture](#infield_correctioncapture)
+service. However, other infield correction services can be used without calling the start service first. This service
+can also be used to restart an active infield correction session.
```
apartridge, line 353: This "However" part sounds a bit weird, is it possible to call /compute, /compute_and_write, /remove_last_capture  without having called /start first?

README.md
```diff
@@ -277,6 +278,127 @@ service call. If the camera is not in `Connected` state the driver will attempt
...
+
+Before calling this service, a call to [infield_correction/start](#infield_correctionstart) must have been made first.
+
+After sufficient number of captures, proceed by calling the [infield_correction/compute](#infield_correctioncompute)
+service to compute the camera correction based on the captures gathered so far. To compute the correction and write it
+to camera, proceed by calling the [infield_correction/compute_and_write](#infield_correctioncompute_and_write) service.
+
+The captured data is cleared if the node is stopped, or after a successful call to either of the services
+[infield_correction/start](#infield_correctionstart) or
+[infield_correction/compute_and_write](#infield_correctioncompute_and_write).
```
apartridge, line 370: I think we should enforce the user to call /start again after having  called /compute_and_write (if that is not how it is already)

micragz:
> If this is for hand-eye, then we should consider to make sepearate PR for infield and merge that first since then those on <2.15 can still get infield supported. But I think we can review both features together in this PR anyway.

We're using 2.15 features for both of them actually.

README.md
micragz: We have two different functions in the SDK, one for compute and one for verify, so this was added for completeness. I guess it is somewhat simpler to do verify than that three-step process, but I'm not sure otherwise if we get any different results. If we get the same information from compute, I wouldn't mind removing this. Although it wouldn't match our other APIs.

@torbsorb, could you fill us in on the main differences between these, do we need a separate verify service?

zivid_camera/src/hand_eye_calibration_controller.cpp
```diff
@@ -0,0 +1,524 @@
...
+{
+  tf2::Transform tf2Transform;
+  tf2::fromMsg(pose, tf2Transform);
+  std::array<tf2Scalar, 16> tf2_transform = {};
+  tf2Transform.getOpenGLMatrix(tf2_transform.data());
+
+  auto transform = transposeMatrix(Zivid::Matrix4x4{tf2_transform.begin(), tf2_transform.end()});
+  transform.at(0, 3) = rosLengthToZivid(tf2Transform.getOrigin().x());
+  transform.at(1, 3) = rosLengthToZivid(tf2Transform.getOrigin().y());
+  transform.at(2, 3) = rosLengthToZivid(tf2Transform.getOrigin().z());
```
apartridge, line 199: I see we modify the translation components, what about if the tf2Transform has a scale applied before? Is that possible and would it work since we are here converting to mm? I guess in practice the matrix wouldn't be likely to have a scale.

README.md
micragz: Yes, you can call those. Although it doesn't make sense before calling /capture first, and you would get another error about missing captures. We could add a more specific error about /start, although you would still need to make a capture too.

The main purpose is to make it clear that you can call the non-stateful services without /start, especially useful for /read and /verify.

README.md
micragz: I placed the non-stateful services above /start, so anything below requires /start (directly or indirectly to succeed). Maybe there is a way to make that distinction more clear?

zivid_camera/src/hand_eye_calibration_controller.cpp
micragz: Yeah, we expect the transform to be in units of meter, as that is the ROS convention. I've tried to document this clearly, so it's up to users to obey this. The transform itself can't actually have a scale in the matrix sense, since it is transmitted as a quaternion in ROS.

README.md
micragz: Changed the behavior as suggested: 461926c122ef4028685f2273e5d9957a2f539fbf

zivid_camera/include/zivid_camera/infield_correction_controller.hpp
```diff
@@ -0,0 +1,121 @@
...
+    std::shared_ptr<std_srvs::srv::Trigger::Response> response);
+  void infieldCorrectionCapture(
+    const std::shared_ptr<rmw_request_id_t> request_header,
+    const std::shared_ptr<zivid_interfaces::srv::InfieldCorrectionCapture::Request> request,
+    std::shared_ptr<zivid_interfaces::srv::InfieldCorrectionCapture::Response> response);
+  void infieldCorrectionCompute(
+    const std::shared_ptr<rmw_request_id_t> request_header,
+    const std::shared_ptr<zivid_interfaces::srv::InfieldCorrectionCompute::Request> request,
+    std::shared_ptr<zivid_interfaces::srv::InfieldCorrectionCompute::Response> response);
+  void infieldCorrectionComputeAndWrite(
```
apartridge, line 97: Since the class is named InfieldCorrection, I think we could remove that prefix from all the method names and member variables?

zivid_camera/test/test_infield_correction_controller.cpp
```diff
@@ -0,0 +1,330 @@
...
+    std::tm * tm_info = std::localtime(&seconds_since_epoch);
+    const int year = tm_info->tm_year + 1900;
+    ASSERT_GE(year, 2025);
+    ASSERT_LE(year, 2125);  // This needs to be updated in 100 years.
+  }
+
+  void infieldStartCaptureWrite()
+  {
+    auto start = doStdSrvsTriggerRequest("infield_correction/start");
+    ASSERT_TRUE(start->success);
```
apartridge, line 82: Just for completeness would also be good to check that message is empty

README.md
apartridge: In Studio I see we are calling `Zivid::Experimental::Calibration::verifyCamera(input).localDimensionTrueness() ` on each of the N measurements taken so far and then presenting those to the user.  I see you do the same in `calculateInfieldCorrectionStatistics`

So verifyCamera(input) returns some info that you don't get from computeCameraCorrection(), but we could still expose that data via the /compute and /compute_and_write ROS services.

What I didnt like about this /verify function is mainly that it is a bit unclear if this capture that it does will end up in the set of current captures that is used by the compute/write function. It may be that we should remove this service and instead provide a list of  `Zivid::Experimental::Calibration::verifyCamera(input).localDimensionTrueness() ` for each of the measurements in the /compute functions.

apartridge: There is value in matching the API as well of course (although in retrospect, maybe we would have simplified the API a little bit). I think with the API we wanted to separate the "verify" part from the "correct" part, but when making Studio we mainly wanted to make it one process and then you can choose to correct or just verify at the end of the capture set.

zivid_interfaces/msg/FixedPlacementOfCalibrationObjects.msg
```diff
@@ -0,0 +1,23 @@
...
+# Specifies the fixed placement of a Zivid calibration board for low degrees-of-freedom hand-eye calibration.
+FixedPlacementOfCalibrationBoard calibration_board
+
+
+### For FIDUCIAL_MARKERS type
+
+# The dictionary that describes the appearance of the given markers. Example: "aruco4x4_50".
+string marker_dictionary
+
+# A list describing the fixed placement of fiducial markers.
```
apartridge, line 22: This one only is needed for low-DOF right?

zivid_interfaces/msg/FixedPlacementOfCalibrationObjects.msg
```diff
@@ -0,0 +1,23 @@
...
+
+# The type of calibration objects to use for low degrees-of-freedom (DOF) hand-eye calibration, if any:
+# NONE (default) or CALIBRATION_BOARD or FIDUCIAL_MARKERS
+# If set to NONE, regular (6-DOF) hand-eye calibration is performed.
+uint8 type
+
+
+### For CALIBRATION_BOARD type
+
+# Specifies the fixed placement of a Zivid calibration board for low degrees-of-freedom hand-eye calibration.
```
apartridge, line 13: Consider making this comment even clearer that for 6DOF this is not needed. It may be confusing what "low degrees" is, so clarifying that this is not needed for "regular" hand eye but just low-dof is best.

> Specifies the fixed placement of a Zivid calibration board for low degrees-of-freedom hand-eye calibration. This field is not needed for regular 6-DOF hand-eye calibration.

zivid_interfaces/msg/FixedPlacementOfCalibrationObjects.msg
apartridge: Then it would be good to clarify that even more in the comment. Also see comment above. https://github.com/zivid/zivid-ros/pull/129/files#r2013896667

zivid_samples/src/sample_hand_eye_calibration.cpp
```diff
@@ -0,0 +1,352 @@
...
+int main(int argc, char * argv[])
+{
+  rclcpp::init(argc, argv);
+  auto node = std::make_shared<rclcpp::Node>("sample_hand_eye_calibration");
+
+  const CalibrationConfiguration configuration =
+    get_parameter_enum(*node, "configuration", calibration_configuration_map);
+  const std::vector<int> marker_ids = sanitize_marker_ids(
+    node->get_logger(),
+    node->declare_parameter<std::vector<int64_t>>("marker_ids", {}, read_only_parameter));
```
apartridge, line 337: Minor, could be const auto

zivid_samples/src/sample_hand_eye_calibration.cpp
```diff
@@ -0,0 +1,352 @@
...
+  constexpr int wait_seconds = 5;
+  constexpr int num_successful_captures_target = 6;
+  RCLCPP_INFO_STREAM(
+    node->get_logger(), "--- Starting captures, will proceed until " +
+                          std::to_string(num_successful_captures_target) +
+                          " captures have completed with a detected calibration object ---");
+
+  bool sleep_before_capture = false;
+  for (int num_successful_captures = 0; num_successful_captures < num_successful_captures_target;) {
+    if (sleep_before_capture) {
```
apartridge, line 294: Minor: Could simplify by just checking `num_successful_captures > 0` here and remove the bool

zivid_camera/include/zivid_camera/infield_correction_controller.hpp
micragz: That would leave some functions/members named just "reset" and "start" for example. I think that's a bit unfortunate? Now they are named after their service name.

apartridge: In the zivid_camera.hpp we use `*ServiceHandler(),  fex `cameraInfoSerialNumberServiceHandler()`, can do the same here for consistency

zivid_camera/test/test_infield_correction_controller.cpp
micragz: Fixed f71a3f9

zivid_interfaces/msg/FixedPlacementOfCalibrationObjects.msg
micragz: c43cc01

zivid_interfaces/msg/FixedPlacementOfCalibrationObjects.msg
micragz: Yes, that's right, fixed here: c43cc01

zivid_samples/src/sample_hand_eye_calibration.cpp
micragz: We might have a non-successful capture here, and still want to sleep here on the next iteration. This would cause it to skip sleeping in that situation.

zivid_samples/src/sample_hand_eye_calibration.cpp
```diff
@@ -0,0 +1,352 @@
...
+    }
+  }
+
+  RCLCPP_INFO(node->get_logger(), "--- Captures complete ---");
+}
+
+void perform_hand_eye_calibration(
+  const rclcpp::Node::SharedPtr & node, const CalibrationConfiguration configuration)
+{
+  RCLCPP_INFO(node->get_logger(), "--- Performing hand-eye calibration with captures data ---");
```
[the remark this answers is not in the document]
micragz: Fixed: 3e2d5722741a19fdffa45aad469958a5fcc04911

zivid_samples/src/sample_hand_eye_calibration.cpp
micragz: Fixed: 3e2d5722741a19fdffa45aad469958a5fcc04911

zivid_camera/include/zivid_camera/infield_correction_controller.hpp
micragz: Yeah, that's quite a bit cleaner:
fd22c60
5485c15

zivid_interfaces/msg/FixedPlacementOfCalibrationObjects.msg
apartridge: I think same as above, you also want to add something like "for low degrees-of-freedom hand-eye calibration" to this line. This doscstring doesnt say now it applies to low-DOF, it just says it doesn't apply to regular 6-DOF.

apartridge:
It is possible to spam this button with many clicks and then the camera will capture for a while. Would be ideal if button was disabled until the camera is ready again.
image

apartridge:
The trueness error is reported as close to 100% for some reason, 
image
Comparing same camera in Studio:
image

apartridge:
Also getting this when trying to write it
image
Same scene works fine in Studio. Maybe something with mm vs. m in the point cloud?

README.md
micragz: Removed the verification service now, and updated the sample to show how to use compute for verification: 34b3a8c

micragz:
> It is possible to spam this button with many clicks and then the camera will capture for a while. Would be ideal if button was disabled until the camera is ready again.

Made this change in 0dbb555

zivid_interfaces/srv/InfieldCorrectionCompute.srv
```diff
@@ -0,0 +1,54 @@
...
+# The list of estimated local dimension trueness error for each capture in this infield correction session. The
+# dimension trueness represents the relative deviation between the measured size of the calibration object and the true
+# size of the calibration object, including the effects of any in-field correction stored on the camera at the time of
+# capture. Note that this estimate is local, i.e. only valid for the region of space very close to the calibration
+# object.
+#
+# The value is a fraction (relative trueness error). Multiply by 100 to get trueness error in percent.
+float32[] trueness_errors
+
+# The local dimension trueness error for this capture. See 'trueness_errors' for more information on this measure.
```
apartridge, line 20: this capture -> the latest capture maybe?
Since there isn't a "this" capture when calling the /compute service.

README.md
```diff
@@ -277,6 +278,138 @@ service call. If the camera is not in `Connected` state the driver will attempt
...
+
+Before calling this service, a call to [infield_correction/start](#infield_correctionstart) must have been made first.
+
+This service call will perform a relatively slow but high-quality point cloud capture with the connected camera. Any
+settings applied to the camera in the ROS driver will be ignored for calls to this service, appropriate settings are
+automatically used. The resulting point cloud and color image will be published just like during a normal call to the
+[capture](#capture) service.
+
+After sufficient number of captures, proceed by calling the [infield_correction/compute](#infield_correctioncompute)
+service to compute the camera correction based on the captures gathered so far. To compute the correction and write it
```
apartridge, line 370: I would write `camera correction and camera verification metrics` or something (check how we phrase it in Studio). Since now this service  (compute) is also what you use to verify the camera now.

zivid_interfaces/srv/InfieldCorrectionCompute.srv
micragz: Yeah, I was thinking along the same lines of perhaps dropping these. We could drop the statistics too (maximum & average), and technically even `number_of_captures`. What do you think about that?

apartridge: I would remove just current_trueness_error

apartridge (commented):
Got no more comments except the other threads. Looking good! Please do a round of testing everything before merging, I have tested infield a bit but not HE.

zivid_interfaces/srv/InfieldCorrectionCompute.srv
micragz: Changed here: ee79c8c59cd8f247354a864de6ee436701c9e28a

zivid_interfaces/srv/InfieldCorrectionCompute.srv
micragz: Couldn't really make this fit well here, I've generally added several links from the readme.

micragz:
I believe I've addressed all comments. Tested manually on existing hand-eye data, and the results are matching the ones from SDK. Will do some more manual testing with low-DOF.

apartridge (dismissed):

apartridge (approved):

PR #130, opened 2025-03-26
Add detection for calibration board and fiducial markers

micragz:


README.md
```diff
@@ -225,6 +225,37 @@ capture, see [capture](#capture).
...
+
+The name of the fiducial dictionary must be provided, along with a list of marker IDs. The scene may not need not
+contain all listed markers for a successful detection. For further information on fiducial markers see [this wikipedia
+page](https://en.wikipedia.org/wiki/Fiducial_marker). For more information on ArUco markers specifically, refer to the
+[OpenCV documentation](https://docs.opencv.org/4.x/d5/dae/tutorial_aruco_detection.html).
+
+The returned data from the service includes a member of the type
+[DetectionResultFiducialMarkers.msg](./zivid_interfaces/msg/DetectionResultFiducialMarkers.msg), with details on any
+detected fiducial markers.
+
```
apartridge, line 258: Should state in the descriptions wether or not these services are publishing the capture topics.

README.md
```diff
@@ -533,6 +564,40 @@ ros2 run zivid_samples sample_capture_2d_cpp
...
+Using ros2 run (when `zivid_camera` node is already running):
+
+```bash
+ros2 run zivid_samples sample_capture_and_detect_calibration_board_cpp
+```
+
+### Sample Capture and Detect Markers
+
+This sample performs a capture and detects any fiducial markers in the current scene. It shows how to invoke the
+[capture_and_detect_markers](#capture_and_detect_markers) service and log the detection result
```
apartridge, line 587: missing dot at the end

zivid_camera/src/detector_controller.cpp
```diff
@@ -0,0 +1,119 @@
...
+    zivid_interfaces::srv::CaptureAndDetectCalibrationBoard::Request> /*request*/,
+  std::shared_ptr<zivid_interfaces::srv::CaptureAndDetectCalibrationBoard::Response> response)
+{
+  RCLCPP_INFO_STREAM(node_.get_logger(), __func__);
+
+  runFunctionAndCatchExceptions(
+    [&]() {
+      auto frame = Zivid::Calibration::captureCalibrationBoard(camera_);
+      auto detection = Zivid::Calibration::detectCalibrationBoard(frame);
+      response->detection_result = toZividMsgDetectionResult(detection);
```
apartridge, line 76: Is response->success default true?

micragz: We set this to true before we start, and false if an exception occurs.

README.md
micragz: Added here: 1575fba

README.md
micragz: 1575fba

zivid_camera/src/detector_controller.cpp
apartridge: Ok right, in `runFunctionAndCatchExceptions`, I feel `runFunctionAndCatchExceptions` is a bit too generically named if it also does special things related to a `response`.

micragz: Yeah, it could have been more clear. Hm, `setResponseSuccessAndRunFunctionAndCatchExceptions` doesn't exactly roll off the tongue?

apartridge: Basically this helper runs a lambda and helps to populate a Trigger style response (?), handling exceptions etc.
Maybe
`runAndCatchExceptionsForTriggerRequest` or something

micragz: Yeah, that sounds much better to me. I used the term `Response` instead of `Request`, as that is the type it fills: https://github.com/zivid/zivid-ros/pull/130/commits/fa7af3be6d163bbdd9f43e598a46336c16153cef

zivid_camera/src/detector_controller.cpp
```diff
@@ -0,0 +1,119 @@
...
+      if (
+        response->detection_result.status !=
+        zivid_interfaces::msg::DetectionResultCalibrationBoard::STATUS_OK) {
+        response->success = false;
+        response->message =
+          "Failed to detect calibration board. " + response->detection_result.status_description;
+      }
+      // Publishing the frame also modifies (transforms) the point cloud in-place to scale it. The
+      // above computations assume a non-transformed point cloud, so publish the frame at the end.
```
apartridge, line 85: Could we enforce this also with a check in the code? Since 2.15, we can check the point cloud's transformation matrix is identity before calling the calibration functions. Might even make sense that `captureCalibrationBoard` (inside the SDK) checks that the point cloud is untransformed (not sure if that is valid usecase)?

zivid_camera/src/detector_controller.cpp
apartridge: nice

zivid_camera/src/detector_controller.cpp
micragz: Added a new commit to put in some checks like that: c54aab3

I guess somebody could in principle want to detect calibration objects in a transformed point cloud, if that's what one already has?

apartridge (approved):

PR #141, opened 2025-05-12
Add sample showing how to use the Zivid SDK directly

micragz:
See readme for details.

### DIFF
--- a/zivid_samples/CMakeLists.txt
+++ b/zivid_samples/CMakeLists.txt
@@ -12,11 +12,12 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
 endif()
 
 function(zivid_add_cpp_sample sample_name)
-  cmake_parse_arguments(PARSE_ARGV 1 ARG "" "" "DEPENDENCIES")
+  cmake_parse_arguments(PARSE_ARGV 1 ARG "" "" "DEPENDENCIES;LINK_TARGETS")
   set(target_name "${sample_name}_cpp")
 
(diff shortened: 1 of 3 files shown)


RMPR (approved):

zivid_samples/CMakeLists.txt
```diff
@@ -34,6 +35,9 @@ find_package(tf2 REQUIRED)
 find_package(tf2_ros REQUIRED)
 find_package(tf2_geometry_msgs REQUIRED)
 
+find_package(Zivid 2.15.0 COMPONENTS Core REQUIRED)
```
vawale, line 38: This now makes all samples to require sdk 2.15 or higher, even if sample does not directly use zivid sdk. Maybe its better to have the samples that use both sdk and ros wrapper in a separate project?

vawale (approved):
Just one comment about required version of zivid sdk for samples.

zivid_samples/CMakeLists.txt
micragz: Right, yeah, it's the same minimum as in the `zivid_camera` driver. So if one can build that, then the samples should work too.

I wonder if there is a use case for building the samples separately, from some other place than with `zivid_camera`? Then it could make sense to make it a separate project. Several samples assume that everything is running on the same machine though.

vawale: ah, if the `zivid_camera` driver also needs 2.15 or higher, then this should be okay. I was under the impression that we support building `zivid_camera` driver with older versions of sdk as well.

vawale (approved):

PR #146, opened 2025-05-13
Add `acquisition_done` topic to zivid_camera

torbsorb:
When using the Zivid SDK directly there is information in when the capture API returns. This indicates that the camera is done with acquisitions. At this point it is safe to move the scene (e.g. robot). In ROS the capture service wraps the capture call, but also extraction and processing of the point cloud. The same goes for 2D capture.

Use this topic to control when to, e.g., move the robot.

### DIFF
--- a/README.md
+++ b/README.md
@@ -377,6 +377,13 @@ Each coordinate is a float value. There are no additional padding floats, so poi
 12 bytes (3*4 bytes). The normals are unit vectors. Note that subscribing to this topic
 will cause some additional processing time for calculating the normals.
 
+### acquisition_done
+
+This topic is published as soon as the camera is done with acquisitions. At this point it
+is safe to move the scene (e.g. robot). This comes before a point cloud or image is necessarily
+available. See the article [Optimize robot cycle times](https://support.zivid.com/en/latest/academy/applications/bin-picking/optimize-robot-cycle-times.html)
+for more information.
+
 ## Configuration
 
 The `zivid_camera` node supports both single-acquisition (2D and 3D) and multi-acquisition
--- a/zivid_camera/include/zivid_camera.h
+++ b/zivid_camera/include/zivid_camera.h
@@ -58,6 +58,7 @@ private:
   bool isConnectedServiceHandler(IsConnected::Request& req, IsConnected::Response& res);
   void publishFrame(const Zivid::Frame& frame);
   Zivid::Frame invokeCaptureAndPublishFrame();
+  bool shouldPublishAcquisitionDone() const;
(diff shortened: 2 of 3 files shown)


zivid_camera/msg/AcquisitionDone.msg
```diff
@@ -0,0 +1,2 @@
+# AcquisitionDone.msg
+std_msgs/Header header
```
micragz, line 2: Do we need our own message type, can we just use the header type directly?

README.md
```diff
@@ -377,6 +377,14 @@ Each coordinate is a float value. There are no additional padding floats, so poi
 12 bytes (3*4 bytes). The normals are unit vectors. Note that subscribing to this topic
 will cause some additional processing time for calculating the normals.
 
+### acquisition_done
+
+When using the Zivid SDK directly there is information in when the capture API returns.
+This indicates that the camera is done with acquisitions. At this point it is safe to move
+the scene (e.g. robot). In ROS the capture service wraps the capture call, but also
+extraction and processing of the point cloud. The same goes for 2D capture. Use this topic
+to control when to, e.g., move the robot.
```
micragz, line 386: Some implementation details here that perhaps are not necessary?

Could say that this is published just after acquisition is done, but possibly before the point cloud is processed and capture service call responds. Maybe with a link to some relevant article on KB?

continuous-integration/lint_cpp.sh
```diff
@@ -13,13 +13,11 @@ if [ -z "$cppFiles$hFiles" ]; then
...
+echo "Formatting C++ source files with clang-format"
 clang-format --version || exit $?
 for fileName in $cppFiles $hFiles; do
-    echo $fileName
-    diff $fileName \
-        <(clang-format $fileName) \
-        || exit $?
+    echo "Formatting $fileName"
+    clang-format -i $fileName || exit $?
 done
```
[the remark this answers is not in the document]
torbsorb: doh.. result of my local short hands for adding to the commit. I did not intend to add this. Nice feature for formatting though.

README.md
torbsorb: Sure, in the context of ROS it is superfluous. However, for a user who is porting from pure Zivid SDK it makes sense, in my opinion.

zivid_camera/msg/AcquisitionDone.msg
torbsorb: Yes, and the changes in the PR are now are much smaller. I have tested that it works just fine without the AcquisitionDone.msg wrapper.

johningve:
Is this already implemented in ROS 2?

PR #153, opened 2025-08-11
Take settings into account for camera intrinsics

johningve:
This ensures the correct resolution is taken into account.

Fixes ZIVID-10518


vawale (approved):
This looks like a big fix to me. Not API breaking.

vawale (approved):

PR #154, opened 2025-08-27
Add ability to estimate intrinsics from frame

torbsorb:
Add the parameter `intrinsics_source` to control how the intrinsics are determined.

Add a new sample to demonstrate intrinsics and the new parameter.

See README.md and
[Camera Intrinsics](https://support.zivid.com/en/latest/reference-articles/camera-intrinsics.html#which-camera-intrinsics-should-i-use) for more information.

### DIFF
--- a/zivid_camera/src/zivid_camera.cpp
+++ b/zivid_camera/src/zivid_camera.cpp
@@ -162,6 +162,7 @@ constexpr auto file_camera_path = "file_camera_path";
 constexpr auto frame_id = "frame_id";
 constexpr auto update_firmware_automatically = "update_firmware_automatically";
 constexpr auto color_space = "color_space";
+constexpr auto intrinsics_source = "intrinsics_source";
 }  // namespace ParamNames
 
 ZividCamera::ZividCamera(ros::NodeHandle& nh, ros::NodeHandle& priv)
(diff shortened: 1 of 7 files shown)


zivid_camera/src/zivid_camera.cpp
```diff
@@ -616,7 +629,21 @@ void ZividCamera::publishFrame(const Zivid::Frame& frame)
...
     if (publish_color_img || publish_depth_img || publish_snr_img)
     {
-      const auto intrinsics = Zivid::Experimental::Calibration::intrinsics(camera_);
+      auto intrinsics = [&] {
+        const auto intrinsics_source = intrinsicsSource();
+        switch (intrinsics_source)
+        {
+          case IntrinsicsSource::Camera:
+            ROS_INFO("Using camera intrinsics for publishing images");
+            return Zivid::Experimental::Calibration::intrinsics(camera_, frame.settings());
```
johnio-zivid-admin, line 638: This code doesn't take into account that the settings could have a different 2D vs 3D resolution. On the ROS2 branch, I made a function to check first if 2D settings are defined, and if so uses that instead of the 3D settings: https://github.com/zivid/zivid-ros/pull/153/files

PR #158, opened 2025-11-20
Share Application in ZividCamera

feuerste:
This PR adds the possibility to construct a `ZividCamera` with an already existing `Application` and a custom node name/namespace.

This is helpful for having additional ROS helper nodes around `ZividCamera`, which also require an `Application` that can only be instantiated once.

### DIFF
--- a/zivid_camera/src/zivid_camera.cpp
+++ b/zivid_camera/src/zivid_camera.cpp
@@ -235,8 +235,10 @@ constexpr auto color_space = "color_space";
 constexpr auto intrinsics_source = "intrinsics_source";
 }  // namespace ParamNames
 
-ZividCamera::ZividCamera(const rclcpp::NodeOptions & options)
-: rclcpp::Node{"zivid_camera", options},
+ZividCamera::ZividCamera(
+  const std::string & node_name, const std::string & ns, const rclcpp::NodeOptions & options,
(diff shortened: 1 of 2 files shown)


zivid_camera/src/zivid_camera.cpp
```diff
@@ -424,6 +427,11 @@ ZividCamera::ZividCamera(const rclcpp::NodeOptions & options)
   RCLCPP_INFO(get_logger(), "Zivid camera driver is now ready!");
 }
 
+ZividCamera::ZividCamera(const rclcpp::NodeOptions & options)
+: ZividCamera(
+    /*node_name=*/ "zivid_camera", /*ns=*/ "", /*options=*/ options, /*application=*/ nullptr)
```
johningve: nitpick: Remove the inline comments here.

johningve (dismissed):
Thank you for the PR! The change looks good to me.

zivid_camera/src/zivid_camera.cpp
feuerste: Done.

feuerste:
Thanks @johningve for your review! I addressed your comment and fixed the formatting, so it should be ready now.

johningve (approved):

PR #164, opened 2026-06-03
ReadMe: Add a note to use millimeters in the HE process

TTPinto:


README.md
```diff
@@ -385,6 +385,9 @@ Performs a hand-eye calibration capture. This service takes the robot pose as an
 detects any calibration objects. The resulting point cloud and color image will be published just like during a normal call to the
 [capture](#capture) service.
 
+> [!IMPORTANT]
+> The robot poses must be given in millimeters.
```
SatjaSivcev: I would add "because the Zivid point clouds come in millimeters."

TTPinto: Changed to:  "The robot poses must be given in millimeters because Zivid point clouds are expressed in millimeters."

RMPR (approved):

PR #168, opened 2026-08-13
Support devcontainer on rootless podman

johningve:
Change workspace folder from /home/ws to /ws. Putting it in home seems unorthodox.

Add support for Podman rootless container. This is a separate devcontainer.json due to some differences vs. the docker version. See commit message for details.


johningve:
Tested OK with CLion. Docker version tested on Ubuntu 26.04, podman version tested on Fedora 44.

vawale (approved):

